PR #23, opened 2014-06-04
Fix SDL2 cursor tracking when menu open

hifi:
The cursor didn't move at all, this implements what is necessary for it to work regardless if the mouse is in relative mode or not (in_raw).


jite:
IIRC this broke hud_editor, think you can take a look at fixing that and I'll merge it. Thanks

jite:
Should be fixed by https://github.com/ezQuake/ezquake-source/pull/67

PR #28, opened 2014-07-04
Make all SDL2 mouse input use relative coordinates

qwcream:
This also fixes a bug where some mouse input was being thrown out
when the mouse is using relative coordinates.


jite:
This sound like the same issue as Issue #23 ?

jite:
Should be fixed by https://github.com/ezQuake/ezquake-source/pull/67

PR #35, opened 2015-02-08
Fix f_system on Mac OS X

JosephPecoraro:
Now it looks like:

now

### DIFF
--- a/host.c
+++ b/host.c
@@ -29,6 +29,10 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #include <sys/time.h>
 #include <machine/cpufunc.h>
 #endif
+#ifdef __APPLE__
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#endif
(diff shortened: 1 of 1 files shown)


host.c
```diff
@@ -248,8 +252,26 @@ void SYSINFO_Init(void)
...
+    	SYSINFO_memory = memsize_value;
+
+    char cpu_brand_string[100];
+    size_t cpu_brand_string_len = 100;
+    if (sysctlbyname("machdep.cpu.brand_string", &cpu_brand_string, &cpu_brand_string_len, NULL, 0) != -1)
+        SYSINFO_processor_description = cpu_brand_string;
+
+    mib[0] = CTL_HW;
+    mib[1] = HW_CPU_FREQ;
+    int cpu_frequency_value;
```
JoakimSoderberg: Haven't touched the ezquake code in a long while, but as I remember, the code was written in C89 compatible style, so that variables are declared at the beginning of the scope always. Maybe I remember wrong, I'm just used to writing all my code like that for max portability.

PR #36, opened 2015-02-08
bugfix for os x bundle: properly handle whitespaces

fzwoch:
bugfix: in case of when the app bundle is installed in paths with whitespace in them it would fail. this patch hopefully escapes them correctly.


JoakimSoderberg:
Great!

PR #37, opened 2015-02-12
Changes related to demo playback

meag:
Mostly changes to let me watch/re-record old demos

NQ demos: player colours set correctly, particle trails/effects back in, demo_jump working, can force team skins with help of cl_demoteamplay
QW demos: can now playback from 2.1 (protocol versions 24/25), demo rewinding works
MV demos: cam_pos stops tracking, if in free camera mode, camera position is restored after rewinding
Video capture: Can split .avi files to be stitched together later (workaround for http://sourceforge.net/p/ezquake/bugs/520/), demo_capture_quiet option to disable playback during capture.
Demo browser: plays the file you select when you have two files with valid demo extensions in same directory

### DIFF
--- a/cl_nqdemo.c
+++ b/cl_nqdemo.c
@@ -305,6 +305,19 @@ static void NQD_ParseClientdata (int bits)
 	}
 }
 
+static void NQD_SetSpectatorFlag(player_info_t* player)
+{
+	player->spectator = (cl.teamplay && player->name[0] && player->real_topcolor == 0 && player->real_bottomcolor == 0 && player->frags <= -99);
+}
+
+void NQD_SetSpectatorFlags(void)
+{
+	int i;
+
+	for (i = 0; i < sizeof(cl.players) / sizeof(cl.players[0]); ++i)
+		NQD_SetSpectatorFlag(&cl.players[i]);
+}
+
 /*
 ==================
 NQD_ParseUpdatecolors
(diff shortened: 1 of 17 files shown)


cl_nqdemo.c
```diff
@@ -291,10 +325,37 @@ static void NQD_ParseUpdatecolors (void)
 	// fill in userinfo values
 	top = min(colors & 15, 13);
 	bottom = min((colors >> 4) & 15, 13);
+
+	if (cl_demoteamplay.integer)
+		cl.players[i].spectator = (top == 0 && bottom == 0 && cl.players[i].frags <= -99);
```
JoakimSoderberg: Would be better to have this as an `is_spectator` function... You use the same code earlier as well

JoakimSoderberg:
Looks good to me. How about you @jite ?

jite:
Yeah I'm gonna test it today and hopefully merge.

jite:
Okey I've tested it quickly and it plays back everything fine, however something else has broken. Timedemo doesn't work as intended. timedemo demo1 reports +30000 frames and plays much slower than it should, so does any other demo (should be about 960 frames).

You can also test by creating your own QWD with qwdrecord and then run timedemo on that with this code and without, big difference.

I also noticed a lot more stuttering when playing back qwd's normally with this code than without.

@meag Can you take a look at this and see if you can find the culprit? Thanks

meag:
The culprit was in cl_main/MinPhysFrameTime()... check for mvdplayback that shouldn't have been there.
Have moved spectator-checking logic into a common function, and fixed bug where cls.demotime could be less than demostarttime in NQ demos, breaking democlock.

Issue #32, opened 2014-12-26
os x: mouse cursor stays visible


freder:
hi,

I compiled ezquake from source, which runs fine — except the mouse cursor doesn't disappear in the game. the menu looks fine, though.

unrelatedly, I used to use the version of ezquake that comes with [nQuake](http://nquake.com/), but the latest one doesn't work in os x yosemite. that one – IIRC – comes as an `*.app` that let's you change settings (like resolution, etc.) before you launch the actual game. — is that nQuake-specific or is there also a way to make an `*.app` when compiling ezquake from source? (I couldn't find any information on this).

jite:
While I think the problem is within the SDL2 library in combination with Apple, there are a couple of things you can try. First try to set vid_width 0 and vid_height 0 to force SDL to use the desktop resolution for ingame. If that doesn't help or it's already what you have, then set vid_width/vid_height to the appropriate values (your res) and see if that helps.

If neither of these help, not that I think this will either, but feel free to try and remove line 154 in vid_sdl2.c (SDL_SetCursor(NULL)).

jite:
Are you still experiencing this with latest master? I can't reproduce it

freder:
I just pulled and recompiled, but I still see the mouse cursor.
however: when I switch to another application and then go back into the game, it's gone.

jite:
Okey. Unfortunately I have no luck trying to reproduce this on my gf's iMac (OSX 10.10) :C What versions of OSX do you run? Gfx cards? Anyone of you think you could try to investigate more (code wise) what might be causing the behaviour? Thanks

freder:
![screen shot 2015-03-02 at 10 52 20](https://cloud.githubusercontent.com/assets/724844/6438605/21004e46-c0cb-11e4-98d0-c2b8ea45f88b.png)

jite:
Thanks @freder . Do you use multiple displays or just the iMac?

fzwoch:
![screen shot 2015-03-02 at 11 22 31](https://cloud.githubusercontent.com/assets/439605/6439006/c1a4b64a-c0ce-11e4-8ad3-d765f2e4ea1d.png)

For the record - happens here too. Using dual monitor setup. It looks like the cursor is reappearing without the SDL2 layer noticing. The API claims the cursor is still hidden.

Not sure if this is a SDL2 bug or some OS X quirk. We need to figure out what exactly is triggering this behaviour first.

[...]

PR #39, opened 2015-03-02
OS X fixes for logo and window/mouse behavior

fzwoch:
- fix low resolution icon os osx
- fix other osx window/mouse issues (hopefully)

Potential fix for issue #32 
Alt-Tab acts differently works around some display issue
Icon now looks nice

### DIFF
--- a/vid_sdl2.c
+++ b/vid_sdl2.c
@@ -522,6 +522,11 @@ static void VID_SDL_GL_SetupAttributes(void)
 
 static int VID_SetWindowIcon(SDL_Window *sdl_window)
 {
+#ifdef __APPLE__
+	// on OS X the icon is handled by the app bundle
+	// it actually is higher resolution than the one here.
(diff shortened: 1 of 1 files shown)


vid_sdl2.c
```diff
@@ -562,6 +568,11 @@ static void VID_SDL_Init(void)
 		}
 	}
 
+	SDL_SetHint(SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES, "0");
```
JoakimSoderberg: Shouldn't this be inside of the `#ifdef __APPLE__` as well? I guess it is just ignored on other platforms...

jite: Yes it should, I put a TODO on it, I can't commit from work so I asked @fzwoch to do the patch. However that line went on the wrong side of the ifdef and he was too lazy to fix it :)

PR #44, opened 2015-10-03
Fixes for cl_demospeed 0 affecting QTV playback

meag:
Particle effects & sounds were paused, but entity movement unaffected.
It was possible previously to pause QTV playback and then start again (as long as qtv_adjustbuffers was set to 0) - if this is still required, can set qtv_allow_pause to 1.  Hopefully the new default causes less confusion.


wernerml:
Luxury!

PR #46, opened 2015-10-07
Bugfix for teamlock setting

meag:
If teamlock is set, the current tracked player may not be a teammate.  Existing code always considers the tracked player a teammate, so +showscores looks confusing when using teamcolor/enemycolor settings.

Example of bugged +showscores here:

ezquake011


wernerml:
GUYS how do I get a Recently build of this amazing bugfixes/improvements ?

I'm interested in this bug-fix here (it is very annoying)

meag:
If you get the latest nightly build from here: http://uttergrottan.localghost.net/ezquake/dev/nightlybuilds/win32/, this change should be included.

PR #48, opened 2015-10-13
Multi-monitors, non-american keyboard layouts, cyrillic characters & a fakeshaft experiment

meag:
This will need a reasonable amount of testing by people with varying configurations.

- Extra cvars to choose which display to use (vid_displayNumber / vid_win_displayNumber).

- Support for keyboard layouts - in SDL2 this means picking up the TextInput event, as the unicode character(s) don't get passed with key up/down.  Have done some testing of other layouts, generating cyrillic characters, euro symbols etc.  See below for change to bind command.

- Cyrillic characters - now support various encoding methods for unicode for outgoing characters (cl_textencoding cvar), but default is 2.2's k8 encoding (more work to be done for utf8 generally)

- Character sets can now cover the 3-byte range of utf8 values (0-0xFFFF), if initial charset is called XXX then system will attempt to load XXX-yyy (yyy = 001 upwards), with special case for -004 falling back to -cyr, for backwards compatibility with existing configurations.

- cl_fakeshaft 3 - very experimental change to 'fix' the shaft's position being antilagged but on/off status and all damage indicators being lagged as usual - probably no benefit over LAN, possibly no benefit at all.  Not expected to hurt nor help LG % very much, but remove worries where the player appears to be shooting the opponent but no damage/blood is appearing.  Downside is an FPS hit for the collision detection of beam & opponents.

- con_bindphysical cvar: bind command behaviour changes if executed in a script or at the console.  Example of problem:

> User is in DVORAK layout
> Type type "bind o +back" at the console.
> To type the o, they would have pressed the button that Quake/QWERTY considers to be S
> They then execute a configuration someone has given them, that person has QWERTY keyboard
> Configuration contains "bind s +back"
> In this case, s means Quake key S

To solve this, con_bindphysical determines if the bind key refers to the physical key or a logical key that matches the user's keyboard layout.  When a configuration file is executed, con_bindphysical is set to 1, and reset back to its original value after execution.  This means that all configuration files passed around will be saved in the second format, and so the layout of the keys for playing the game will remain constant.

If there was a need for a key to correspond to a command (for example, a demo-watching config where T toggled teaminfo overlay, R toggled layout etc) then the configuration file would have to contain "con_bindphysical 0" at the top... then whoever executed it (on whatever layout) would always press the key labelled on their keyboard as R, regardless of what layout it was.  Hopefully this keeps backwards compatibility with existing config files but allows an easy way to bind keys at the console.

### DIFF
--- a/cl_main.c
+++ b/cl_main.c
@@ -128,6 +128,8 @@ cvar_t	cl_demospeed = {"cl_demospeed", "1"};
 cvar_t	cl_staticsounds = {"cl_staticSounds", "1"};
 cvar_t	cl_fakeshaft = {"cl_fakeshaft", "0", 0, Rulesets_OnChange_cl_fakeshaft};
 cvar_t	cl_fakeshaft_extra_updates = {"cl_fakeshaft_extra_updates", "1"};
+cvar_t  cl_fakeshaft_clientside_beams = {"cl_fakeshaft_clientside_beams", "1"};
+cvar_t  cl_fakeshaft_clientside_damage = { "cl_fakeshaft_clientside_damage", "1" };
 cvar_t	cl_parseWhiteText = {"cl_parseWhiteText", "1"};
 cvar_t	cl_filterdrawviewmodel = {"cl_filterdrawviewmodel", "0"};
 cvar_t	cl_demoPingInterval = {"cl_demoPingInterval", "5"};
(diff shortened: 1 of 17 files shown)


JoakimSoderberg:
Looks nice :)

PR #54, opened 2015-10-15
Don't download mingw32-libs from external source

anderslindmark:
Now that the mingw32-libs are in the repository we don't need to download them when building the image

### DIFF
--- a/misc/docker-nightly-build/Dockerfile
+++ b/misc/docker-nightly-build/Dockerfile
@@ -7,8 +7,7 @@ RUN pacman-key --init && pacman-key --refresh-keys
 RUN pacman --noconfirm -Syu && pacman-db-upgrade
 RUN pacman --noconfirm -S gcc git make pkg-config p7zip sdl2 libpng jansson libjpeg-turbo mingw-w64-gcc
 
-# Add libraries and buildscript
-ADD mingw32-libs.tar.gz /usr/i686-w64-mingw32
+# Add buildscript
 ADD mk_nightly.sh /
 
 # Target volume for the nightly archive
--- a/misc/docker-nightly-build/README.md
+++ b/misc/docker-nightly-build/README.md
@@ -2,7 +2,11 @@ This method of building ezQuake requires *Docker*. You can read more about it an
 
 Initialization
 ==============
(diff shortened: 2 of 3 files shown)


JoakimSoderberg:
@jite ouch! would've been preferable with binary files like that in a separate repo and in a submodule instead. Those Megabytes will be there even if they're deleted in the future

JoakimSoderberg:
@jite I can fix it in a nicer way, however that will require me to rewrite some of the history (last couple of commits)... tell me if you think that's a good idea.

jite:
@JoakimSoderberg I'd been thinking about it for a while and couldn't really come up with any drawbacks else than it takes about an extra 5MB download upon for clone. (The current git history is in the 50MB region). I don't see these extra 5MB (compressed size) as a problem from a size perspective.

I also made it like this out of pure convenience, if not only for me but for the people using it. We only link libraries statically for the Windows platform. If this were to change in the future I think it could be reasonable having them in another repo, just like it was previously when all platforms linked statically.

This is how my thoughts go anyway, but I'm all ears though if you have good reasons to change it :) Any major drawbacks doing it this way in your opinion (except for the space) or any benefits of doing it the other way?

JoakimSoderberg:
well I mean the exact same setup except that instead of a directory with the libs it is a git submodule. It won't add size to the main repo, and people not compiling for windows don't need to fetch it. Any binary bloat is bad in my opinion :)

With submodules:
```
git clone --recursive <url>
```

Normal:
```
git clone <url>
```

All other aspects are the same. Except when updating the libs that is, but that's not very often.

Aslo, it's nicer to put all those instructions in the nicely formatted [`README.md`](README.md) so that github can format and display it on the repo page nicely.

jite:
I tried to weight the benefit of simplicity for our mostly "not-that-technical" users in qw community vs the binary bloat. My conclusion was that this was the simplest/easiest (for now).

I am thinking of a complete repository make over/restructure in the future. That also goes for code etc since the repo right now is a rather plain one directory that contains it all.

README.md would be nice, pull requests are happily accepted. It's on the TODO list for future improvement. Feel free to join up on IRC again, much easier to talk :)

PR #63, opened 2015-12-09
JSONHELP: Convert xml documentation to json

meag:
.json file embedded in executable, loaded on startup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ STRIP ?= strip
 RM ?= rm -f
 RMDIR ?= rm -rf
 MKDIR ?= mkdir -p
+XXD ?= xxd -i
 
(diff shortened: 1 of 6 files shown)


johnnycz:
There's code in ezquake-documentation that reads the XML files and generates most of the documentation website content out of it. If you drop XML generation support and at the same time not update ezquake-documentation, it will no longer be possible to update ezQuake's documentation website automatically.

jite:
Yeah. The main purpose is to have the cvar/cmd documentation in the repo, tightly coupled with the src code and then built in to the client. Right now there's a mess where many people don't have any documentation or they have mismatching docs and clients (not mentioning that the docs aren't up to date :))

We're working on replacing everything sourceforge related, don't want anything to do with them. The web manual will most likely end up here at github and being generated from the json files.

PR #66, opened 2015-12-27
.mvd support for OS X

fzwoch:
These changes add .mvd file support to OS X. You double click a .mvd file from the finder and the app launches and plays the demo. Or if it is already launched it will play the demo that has been clicked. You can also drag and drop the file onto the app's icon of course.


jite:
Thanks

PR #84, opened 2016-04-15
HELP: Correcting errors & adding missing data

meag:
Fix error made during import where some enums were converted to strings.
If type is missing, no longer reports "boolean value" (reported by Ch0wW)
Tried to add some more descriptions/types to missing variables
Added /help_issues command to list errors in documentation.


meag:
Deleting.. need to not trust Visual Studio... :)

PR #89, opened 2016-05-25
common.c: allow multiple GPL map hashes

fzwoch:
This change allows the GPL map hash table to include multiple entries for one map. This way we can support different GPL map versions. (I'm playing around with this currently)

Also currently "end" is being listed twice. I think the second entry is actually ineffective.


meag:
Closing: commit https://github.com/ezQuake/ezquake-source/commit/fc14070360d1e414807dd39d690dcbc71b25f95e

PR #110, opened 2016-07-09
LEAKS: Fix leaks during console tab completion

JosephPecoraro:
Numerous leaks when doing tab completion in the console.

Namely every time either escape_regex or str_repeat is called, if they allocated they will leak.

To reproduce:

1. Open console
2. Type "src" <tab>

### DIFF
--- a/keys.c
+++ b/keys.c
@@ -309,11 +309,14 @@ void PaddedPrint (char *s)
 {
 	extern int con_linewidth;
 	extern char *str_repeat (char *str, int amount);
+	char *pad = NULL;
 	int nextcolx = 0;
 
 	if (con_completion_format.integer)  // plain list
 	{
-		Com_Printf ("%s&c%s%s&r\n", str_repeat(" ",con_completion_padding.integer), con_completion_color_name.string, s);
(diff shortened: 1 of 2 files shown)


jite:
Good finds, there are a lot of these "stupid" leaks everywhere. I'd recommend that when touching code that uses va(), rewrite it to get rid of the va() usage instead. I don't like assignments within func/macro calls. Would you mind rewriting it a bit to get rid of the va() usage completely? Thanks

jite:
Oh and btw, hello from sunny honeymoon in Skiathos, Greece :)

JosephPecoraro:
Hello ;)

Good suggestions. I made a helper function to print the mark and pad. It makes the `str_repeat` allocation tracking simpler, and simplifies `PaddedPrintValue` by avoiding crazy long lines.

meag:
Included: https://github.com/ezQuake/ezquake-source/commit/a3e790ba8a3bf5a2f7c47814d41f18602b2b48a6 & https://github.com/ezQuake/ezquake-source/commit/2c07a85540220534a998c33d344a4d271a0c0eab

PR #115, opened 2016-07-11
ALL: Fix gcc warnings about broken indentation

jite:
Indentation still far from perfect but hopefully a bit better now in
these files


meag:
Closing: commit https://github.com/ezQuake/ezquake-source/commit/dd1fa9bc790c9484de826098392a73a05cd21e68

PR #124, opened 2016-07-19
GL_RMAIN: Extend custom LG color to all projectiles

### LINKED ISSUE #123: REQ: Introduce ability to set color of all projectiles
This was introduced in 3.0 to allow the LG beam color to be set without the need to replace textures.

Request is to allow all projectiles to have equivalent functionality.
Suggested by dev on Discord.

meag:
gl_custom_X_color is a color cvar that dictates the color of the projectile
  (if specified, alpha is ignored)
gl_custom_X_fullbright is boolean to dictate if the projectile should
  be drawn fullbright, or lit & textured as normal.

Fixes #123.


meag:
Closing: commit https://github.com/ezQuake/ezquake-source/commit/39725f3a7c42718cb34ee69981eb5d3ded14659c

PR #151, opened 2016-08-21
Committer: Vladimir Zhuravlev <mmavova@gmail.com>

mmavova:
modified:   cl_demo.c

Closes ezQuake/ezquake-source#148.


mmavova:
dimman has a better fix.

PR #162, opened 2016-08-30
Fixes ezQuake/ezquake-source#152.

mmavova:
Fixes ezQuake/ezquake-source#152.

### DIFF
--- a/fs.c
+++ b/fs.c
@@ -1626,7 +1626,14 @@ int FS_ZipUnpackOneFileToTemp (unzFile zip_file,
 
 	if (retval == UNZ_OK) {
 		char tmp_path[MAX_OSPATH];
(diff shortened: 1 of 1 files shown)


fs.c
```diff
@@ -1626,7 +1626,14 @@ int FS_ZipUnpackOneFileToTemp (unzFile zip_file,
 
 	if (retval == UNZ_OK) {
 		char tmp_path[MAX_OSPATH];
-		snprintf(&tmp_path[0], sizeof(tmp_path), "%s%s", unpack_path, filename_inzip);
+                if (keep_path)
```
jite: Use tabs for indentation, not spaces. Also make it look like:

if () {
} else {
}
style wise.

If you update with the changes I'll pull it in. Thanks

PR #163, opened 2016-08-30
Fixes ezQuake/ezquake-source#150.

mmavova:
Fixes ezQuake/ezquake-source#150.


jite:
Fixed this in: 17e9a268053b326e54fc04bc76e7093eed5a8a34

Got rid of preceeding __ to using std inline. Thanks

Issue #172, opened 2016-09-06
BUG: crash on lgc3 


mmavova:
IF I run lgc3 ezquake crashes with the following error message:
Error: VFSOS_ReadBytes: bytestoread < 0

steps to reproduce:
1. type "gamedir lgc3" in console
2. type "exec lgc3"
Kaboom!
Error: VFSOS_ReadBytes: bytestoread < 0

http://lgc.quakeworld.nu/?action=about

Matrix-QW:
Works for me.

mmavova:
What OS are you on?

On 09/08/2016 07:06 AM, Matrix-QW wrote:
> Works for me.
>
> — You are receiving this because you authored the
> thread. Reply to this email directly, view it on GitHub
> <https://github.com/ezQuake/ezquake-source/issues/172#issuecomment-245565023>,
>
>
or mute the thread

Matrix-QW:
Windows 8.1 64 bits

mmavova:
Here is what I see in gdb:
Thread 1 "ezquake-linux-x" hit Breakpoint 1, VFSOS_ReadBytes (file=0x5a39690, buffer=0x7fffeab35820, bytestoread=64008, err=0x7fffffffd27c) at vfs_os.c:35
35		if (bytestoread < 0)
(gdb) bt
#0  VFSOS_ReadBytes (file=0x5a39690, buffer=0x7fffeab35820, bytestoread=64008, err=0x7fffffffd27c) at vfs_os.c:35
#1  0x0000000000417062 in FS_LoadFile (path=path@entry=0x57c535 "gfx/conback.lmp", usehunk=usehunk@entry=1, file_length=file_length@entry=0x0) at fs.c:390
#2  0x000000000041711d in FS_LoadHunkFile (path=path@entry=0x57c535 "gfx/conback.lmp", len=len@entry=0x0) at fs.c:399
#3  0x0000000000522259 in Draw_InitConback () at gl_draw.c:1927
#4  0x0000000000417e15 in FS_SetGamedir (dir=dir@entry=0x8aa770 <cmd_tokenizecontext+656> "lgc3") at fs.c:629
#5  0x0000000000436277 in SV_Gamedir_f () at sv_ccmds.c:1760
#6  0x000000000040cafc in Cmd_ExecuteStringEx (context=context@entry=0xeeb180 <cbuf_main>, text=text@entry=0x7fffffffdc80 "gamedir lgc3") at cmd.c:1719
[...]

PR #174, opened 2016-09-10
Fixes ezQuake/ezquake-source#172

mradrian001:
Fixes issue #172 
When loading the gamedir, at some point it tries to run fopen(...) on a directory. In linux, running fopen(...) on a directory returns a valid file descriptor but with a file length of -1 (this is what causes problems). 

For more info:
http://stackoverflow.com/questions/18192998/plain-c-opening-a-directory-with-fopen

### DIFF
--- a/fs.c
+++ b/fs.c
@@ -353,6 +353,8 @@ static byte *FS_LoadFile (const char *path, int usehunk, int *file_length)
 	if (!f)
 		return NULL;
 	len = VFS_GETLEN(f);
+	if(len == -1)
+		return NULL;
 	if (file_length)
 		*file_length = len;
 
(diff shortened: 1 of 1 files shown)


fs.c
```diff
@@ -353,6 +353,8 @@ static byte *FS_LoadFile (const char *path, int usehunk, int *file_length)
 	if (!f)
 		return NULL;
 	len = VFS_GETLEN(f);
+	if(len == -1)
+		return NULL;
```
qqshka: Probably VFS_CLOSE(f) required.

mradrian001: @qqshka Thanks!

mmavova:
I am confused. It does not work for me with the built-in server.
It loads everything, but "addbot" does nothing. Weird, huh?

meag:
This was squashed & included here: https://github.com/ezQuake/ezquake-source/commit/de2ec5fb51af335a0552d79421553b5e1345f1a0

PR #180, opened 2016-10-29
fix fixbundle.sh

fzwoch:
There is no point in parsing Info.plist. In our case the executable is a shell script and no binary. Newer Xcode versions throw and error here.


fzwoch:
ugh.. it may need some more careful changes.. hold on before merging.

PR #187, opened 2017-01-04
Edit some cvars infos (/describe)

ViantriZ:
`cl_clock 1` - it's no longer displayed in status bar when pressing TAB.
`hud_score_bar_fixed_order 1` https://github.com/ezQuake/ezquake-source/issues/184 >> %t should be the current point of view, not the leading team - is this not happening? - @meag
`hud_frags_style 3` - similar to 1? more like 0
`hud_teamfrags_style 3` - same as above
`scr_compactHud 4` doesn't seem to show compact hud with weapon's ammo in big numbers anymore

And that all what I remember.

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -603,7 +603,7 @@
       "type": "enum",
       "values": [
         { "name": "0", "description": "Off." },
-        { "name": "1", "description": "Shows the time you spent on server (you'll also see it in the status bar when you press TAB)." },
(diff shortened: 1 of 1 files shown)


meag:
Thanks for this - looks like hud_frags_style 3 is identical to 0, maybe a mistake when creating 4 and up.
Looks like the Travis checks are failing because of changes to osx images, unrelated to this change.

Thanks again

PR #195, opened 2017-01-18
Edit some cvar infos (/describe)

ViantriZ:
~~What? Again? Yes.~~
`hud_gun_style 4` - duplicated string from `hud_gun_style 1`. What? Maybe, "Number, gold"?
`hud_gun*_style` `6` and `7` - someone accidentally swapped these strings or an mistake when creating 5 and up?


meag:
Accepted https://github.com/ezQuake/ezquake-source/commit/5aa7c681746bd0c97a1672cd46508114683c259f

meag:
They're actually the same in the background, we should probably allow the json help file to reference a common set of values.

PR #213, opened 2017-04-12
HUD: do not obfuscate weapons/ammo in compact hud (4) while tracking

ciscon:


ciscon:
added fix for lto/o3+ compilation/linking

Makefile
```diff
@@ -56,7 +56,7 @@ else
 
     # Resolve all symbols at link time
     ifeq ($(SYS),Linux)
-    	CFLAGS += -DX11_GAMMA_WORKAROUND
+        CFLAGS += -DX11_GAMMA_WORKAROUND -fno-tree-vectorize
```
jite: Please keep this in a separate commit. Not sure we should merge it tbh. What you could do it keep a local .config file where you add it to your CFLAGS for builds with higher optimization levels. (O3 is rarely recommended due to many potential bugs, is there a reason to why you want to compile with O3?)

vx_stuff.c
```diff
@@ -91,7 +91,7 @@ cvar_t		amf_part_teleport = {"gl_particle_telesplash", "0"};
 cvar_t		amf_part_fire = {"gl_particle_fire", "0"}; // 1
 cvar_t		amf_part_firecolor = {"gl_particle_firecolor", ""}; // 1
 cvar_t		amf_part_sparks = {"gl_particle_sparks", "0"}; // 1
-cvar_t		amf_part_traillen    = {"gl_particle_trail_lenght", "1"};
+cvar_t		amf_part_traillen    = {"gl_particle_trail_length", "1"};
```
jite: Could you make this (and the corresponding doc change) a separate commit too please. This will break configs as current configs have the misspelled cvar stored so it should be easy to spot in the git history. Thanks

gl_rmain.c
```diff
@@ -217,6 +218,8 @@ cvar_t gl_modulate                         = {"gl_modulate", "1"};
 cvar_t gl_outline                          = {"gl_outline", "0"};
 cvar_t gl_outline_width                    = {"gl_outline_width", "2"};
 
+#define ISDEAD(i) ( (i) >= 41 && (i) <= 102 )
```
jite: There's already 3 defines for this macro, lets not create a fourth :)

Makefile
ciscon: tree vectorization doesn't get enabled until O3 anyway so unless someone explicitly sets O3 on it shouldn't change anything.  I do get a small (somewhere around 5%) speed increase at O3, and this is the only optimization enabled by it that breaks anything.

gl_rmain.c
ciscon: Understandable, how would you like me to handle it?

ciscon:
separated out commits

PR #215, opened 2017-04-28
add pthread on windows

heeen:
fixes missing symbol sched_yield from libjansson

mingw32-libs/lib/libjansson.a(hashtable_seed.o):(.text+0x41): undefined reference to `sched_yield'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -332,7 +332,7 @@ ifdef CONFIG_WINDOWS
 	localtime_win.o \
 	sys_win.o \
 	winquake.o
-    LIBS_c += -lopengl32 -lws2_32 -lwinmm
+    LIBS_c += -lopengl32 -lws2_32 -lwinmm -lpthread
 else
     OBJS_c += \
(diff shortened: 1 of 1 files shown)


heeen:
By the way, 2.2 does not launch for me on windows, no error message or anything on the console. Building with this works for me

jite:
On what platform are you building and how are you building it (crosscompilation for Windows that is)?

meag:
(Can confirm that this is problem when compiling on Windows with MinGW/MSYS, and the suggested fix works here)

heeen:
win 10/linux subsystem with mingw crosscompiling to windows native.

jite:
Ok, thanks! While not "officially" supported (compiling on windows) it's great if we can make it work :)

heeen:
compiling on lxss is more like compiling on linux given that it emulates the kernel interface and runs ELF binaries etc. It might be an ubuntu issue?
thanks for the merge

PR #225, opened 2017-07-18
Update COMPILING_ON_LINUX.txt

adem4ik:
- Removed 'gitk' from dependicies, it is not required for cloning or building
- Added dependecies for Ubuntu 16.10+ (libpng12-dev was replaced by libpng-dev)
- Added some more steps


jite:
Thanks! If you don't mind feel free to update README.md with the same info, plan is to remove these custom files and just keep the info in the readme. Thanks again for helping out

PR #231, opened 2017-08-28
added fedora dependencies to linux compilation readme

ciscon:


ciscon:
Added new commit to fix issue with match demo autorecord when user's homedir is not on the same filesystem as gamedir (*nix)

PR #234, opened 2017-09-13
Added support for ball-based TF maps

davecenteno:
Added 3 lines for "dropped the ball" "picked up the ball" and "scores" function on non-Spaz ball-based maps.


meag:
Sorry - know we discussed this briefly before, but when editing in github web, it warns "We've detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8." ... I think the ">128 = brown characters" quake charset is confusing it.

Can you edit in a standard text editor and try again?  Hopefully then just your added lines will show in the diff.

Thanks,
meag

PR #237, opened 2017-10-08
BUG-161: Remove temp files after demo playback.

DavidWiberg:
Remove the temporary demo file, which is created during playback from archive, when the demo has been loaded in memory (see https://github.com/ezQuake/ezquake-source/issues/161).

### DIFF
--- a/cl_demo.c
+++ b/cl_demo.c
@@ -3518,6 +3518,7 @@ void CL_Play_f (void)
 	#ifndef WITH_VFS_ARCHIVE_LOADING
 	#ifdef WITH_ZIP
 	char unpacked_path[MAX_OSPATH];
+	qbool extracted_from_archive = false;
 	#endif // WITH_ZIP
 	#endif // WITH_VFS_ARCHIVE_LOADING
 
@@ -3553,6 +3554,7 @@ void CL_Play_f (void)
 	if (CL_GetUnpackedDemoPath (Cmd_Argv(1), unpacked_path, sizeof(unpacked_path)))
 	{
 		real_name = unpacked_path;
+		extracted_from_archive = true;
 	}
 	#endif // WITH_ZIP
 	#endif // WITH_VFS_ARCHIVE_LOADING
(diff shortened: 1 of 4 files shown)


cl_demo.c
```diff
@@ -3665,6 +3667,19 @@ void CL_Play_f (void)
...
 			VFS_CLOSE(playbackfile);
 			playbackfile = mmap_file;
+			#ifdef WITH_ZIP
+			if (extracted_from_archive)
+			{
+				char tmp_dir[MAX_OSPATH];
+				if (COM_StripFilename(name, tmp_dir, sizeof(tmp_dir)))
+				{
+					// Remove the extracted file
+					unlink(name);
```
jite: Sys_remove

cl_demo.c
```diff
@@ -3665,6 +3667,19 @@ void CL_Play_f (void)
 			// Close the file on disk now that we have read the file into memory
 			VFS_CLOSE(playbackfile);
 			playbackfile = mmap_file;
+			#ifdef WITH_ZIP
+			if (extracted_from_archive)
+			{
+				char tmp_dir[MAX_OSPATH];
+				if (COM_StripFilename(name, tmp_dir, sizeof(tmp_dir)))
```
jite: Please change to return -1 on error and 0 on success and perhaps rename to COM_DirName?

Also, some extra sanity checks ain't wrong to make sure you don't try to remove '/' or 'c:/' etc by mistake

common.c
```diff
@@ -122,6 +122,38 @@ char *COM_SkipPathWritable (char *pathname)
 	return last;
 }
 
+int COM_StripFilename(const char *pathname, char *dir, int dir_size)
```
jite: how about: full_path, out_buf, out_size

common.c
```diff
@@ -122,6 +122,38 @@ char *COM_SkipPathWritable (char *pathname)
...
+		i++;
+	}
+
+	if ((index_last_sep == -1) || (index_last_sep >= dir_size))
+	{
+		// No directory separator found or buffer too small.
+		return 0;
+	}
+
+	strncpy(dir, pathname, index_last_sep);
```
jite: Please use `strlcpy` instead

meag:
Should be fixed by https://github.com/ezQuake/ezquake-source/commit/d02add0d31e49e713f758305ca0559856c9186e2

PR #238, opened 2017-10-08
Fix return value inconsistencies for temp files.

DavidWiberg:
The functions "COM_GetUniqueTempFilename" and "COM_GetTempDir" contained
inconsistent handling of return values in some error cases (errors were
being masked).

I noted that error handling strategy varies throughout the code but I opted for zero on failure as that was how the functions were being used.

### DIFF
--- a/fs.c
+++ b/fs.c
@@ -1380,7 +1380,7 @@ int FS_GZipUnpackToTemp (char *source_path,		// The compressed source file.
 {
 	int retval;
 	// Get a unique temp filename.
-	if (!COM_GetUniqueTempFilename (NULL, unpack_path, unpack_path_size, true))
+	if (COM_GetUniqueTempFilename (NULL, unpack_path, unpack_path_size, true) < 0)
 	{
 		return 0;
 	}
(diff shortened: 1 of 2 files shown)


DavidWiberg:
Thanks for the feedback!

I have updated the branch to fix the cases where I spotted errors during check of return values.

fs.c
```diff
@@ -1792,7 +1792,7 @@ int FS_ZipUnpackToTemp (unzFile zip_file,
 	int	retval = UNZ_OK;
 
 	// Get a unique temp filename.
-	if (!COM_GetUniqueTempFilename (NULL, unpack_path, unpack_path_size, true))
+	if (!COM_GetUniqueTempFilename (NULL, unpack_path, unpack_path_size, true) < 0)
```
JoakimSoderberg: `COM_GetUniqueTempFilename` returns `-1` on error and `strlen(filename)` otherwise.

Get rid of the `!` here. `!(-1)` becomes `0`. Then you compare `0 < 0`. So you never enter this `if` statement.

DavidWiberg: Absolutely true! For some reason I only managed to do that to one call site.

Code has been updated.

jite:
Thanks

PR #239, opened 2017-10-10
Mainmenu aspect ratio

jonaslinn:
As I have a 21:9 monitor, the menu is way too stretched.
Before:
ezquake002
After:
ezquake003

I also took portrait aspect ratio into consideration, so the menu will still look good if the height is bigger than the width.

I am not sure about vid.aspect, it is defined in vid.h viddef_t. But as far as I can see never actually set.

scr_scaleMenu is now the scale factor of the menu.

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -12587,10 +12587,10 @@
     },
     "scr_scaleMenu": {
       "group-id": "17",
-      "type": "boolean",
+      "type": "float",
+      "desc": "Scales the frontend menu.",
       "values": [
-        { "name": "false", "description": "Disable scaling." },
-        { "name": "true", "description": "Scales the frontend menu to match the resolution you are using." }
+        { "name": "0", "description": "Disable scaling." },
       ]
     },
     "scr_scoreboard_afk": {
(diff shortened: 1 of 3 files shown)


ciscon:
vid_conwidth/height/scale should take care of this.

jonaslinn:
Unfortunately they don't.

Yes, vid.height depends on conwidth / conheight. But as you can see in the code, the menuwidth was hardcoded to be 320 regardless of the aspect ratio. So the menu will always be stretched over the full width.

ciscon:
scr_scalemenu

jonaslinn:
Yes, with `scr_scalemenu 0` it isn't stretched, but rather small. only with something like `vid_conscale 5` it gets an ok size. But this puts every hud in complete meltdown.

As this menu is the first thing a player sees, i thought it would be nice if it isn't stretched or not tiny.

ciscon:
i see your point now, i was just thinking you were going for proper aspect ratio and didn't care whether or not it was scaled at all (which is what i do).

jite:
Thanks and sorry for late action.

PR #243, opened 2018-01-26
BUILD: Added a simple build script for linux users (debian/ubuntu/fed…

ciscon:


Calinou:
Note that ``command`` is deprecated syntax for `$(command)`. The latter should probably be used everywhere.

[ShellCheck](https://www.shellcheck.net/) helps a lot in writing clean shell scripts. :smiley:

build-linux.sh
```diff
@@ -0,0 +1,98 @@
...
+	echo "You must be root to run this script as it interfaces with your package manager, run with sudo or root privileges."
+	exit 1
+fi
+
+if [ ! -e Makefile ];then
+	echo "Cannot find Makefile, please run this script from the source-code directory."
+	exit 1
+fi
+
+function cpu_cores() {
```
Calinou: You can call `nproc` to retrieve the number of CPU threads. Consider building with `(($(nproc) * 3/2))` threads (150% of CPU threads, which is a bit faster). The operation looks a bit funny (and it only works in Bash), but that's because Bash does not feature floating-point arithmetic :smile:

build-linux.sh
```diff
@@ -0,0 +1,98 @@
...
+	old=$(grep -e 'stretch' -e '8.[0-9]' -c '/etc/debian_version')
+elif [ ! -e /etc/redhat-release ];then
+	echo "Your environment isn't supported by this script.  Bailing out."
+	exit 1
+fi
+
+if [ $debian -eq 1 ];then
+
+	echo "Updating apt repo list..."
+	apt update -y -qqq
```
Calinou: `apt` isn't recommended for usage in scripts (as its interface is not stable), stick to `apt-get` for scripts.

build-linux.sh
```diff
@@ -0,0 +1,98 @@
...
+	check_status
+
+	if [ $old -eq 1 ];then
+		packages="$deb_old"
+	else
+		packages="$deb"
+	fi
+
+	echo "Installing packages..."
+	apt install -y -qqq $packages
```
Calinou: Should installation logs be quieted? I don't think it's that problematic to leave them untouched, especially since dangerous stuff may happen (such as a package being removed due to a conflict).

build-linux.sh
```diff
@@ -0,0 +1,98 @@
...
+
+trap "if [ \$? -ne 0 ];then echo -e \"${RED}Failure, bailing out.${NC}\";if [ -e $buildlog ];then echo \"Please see $buildlog for details.\";fi;exit 0;fi" INT TERM EXIT
+
+if [ `id -u` -ne 0 ];then
+	echo "You must be root to run this script as it interfaces with your package manager, run with sudo or root privileges."
+	exit 1
+fi
+
+if [ ! -e Makefile ];then
+	echo "Cannot find Makefile, please run this script from the source-code directory."
```
Calinou: Grammar: `source-code` → `source code`

jite:
Hi,
First off, thanks! Secondly, sorry for taking long time to respond. I hope to get time to give this some thought this week.

meag:
Merged after rebase: https://github.com/ezQuake/ezquake-source/commit/25d0d505d56652cb9b215014f7c358c55bb254f7

PR #244, opened 2018-01-31
BUILD: remove build dependency on xxd utility

qirex-rd:
The xxd utility is included with Vim.  It is used in the build
process to convert JSON files into a representation suitable for
inclusion in a C program.

The same task can be accomplished with just regular POSIX
utilities -- mainly awk and od -- which are already present
on most operating systems.  This avoids having to install an
additional complete editor just for this small task in the build
process.


fzwoch:
I give an extra kudos for fixing `meson.build` an its helper.

qirex-rd:
Ha, I did forget to test this with `meson`. It's fixed now.

fzwoch:
Oh it didn't work before? I haven't checked either but it looked okay..

qirex-rd:
The generated variable name was different when building with meson, because it included the path to the source file relative to the build directory. So `../` was turned into a `___` prefix, which is of course not what we want.

I also fixed a few other issues along the way.

meag:
Merged after rebase: https://github.com/ezQuake/ezquake-source/commit/01f1554b53d35fd09c7f9406ab7b063140f83ce8

PR #249, opened 2018-03-20
Fixed server loading on a Mac + small refactor of source loading src

krychu:
### Development
- Fixes issue on a Mac where servers are not refreshed properly after updating sources.
- Reworked SB_Update_Source_From_URL. It doesn’t rely on file creation to update servers - so now it will also work on read-only media.
- Refactored SB_Update_Source_From_URL into more focused funcs.
- Additional error checking.

### Testing
- Tested on a Mac Book Pro
- Would be good to test on a Windows / Linux machine ;)

### DIFF
--- a/EX_browser_sources.c
+++ b/EX_browser_sources.c
@@ -305,6 +305,7 @@ struct curl_buf *SB_Retrieve_Data(const source_data *s)
 		return NULL;
 	}
 
+	curl_easy_cleanup(curl);
 	return curl_buf;
 }
 
--- a/EX_browser_sources.c
+++ b/EX_browser_sources.c
@@ -210,9 +210,10 @@ static void SB_Process_URL_Buffer(const struct curl_buf *curl_buf, server_data *
 	char *buf = Q_malloc(sizeof(char) * curl_buf->len);
 	memcpy(buf, curl_buf->ptr, sizeof(char) * curl_buf->len);
 
-	netadr_t addr;
 	// Not using strtok as it's not thread safe
-	for (char *p0 = buf, *p1 = buf; p1 < buf + curl_buf->len; p1++) {
+	netadr_t addr;
+	char *p0, *p1;
+	for (p0 = buf, p1 = buf; p1 < buf + curl_buf->len; p1++) {
 		if (*p1 == '\n') {
 			*p1 = '\0';
 			NET_StringToAdr(p0, &addr);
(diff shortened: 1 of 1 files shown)


EX_browser_sources.c
```diff
@@ -197,66 +204,129 @@ static void Precache_Source(source_data *s)
...
+			p0 = p1 + 1;
+		}
 	}
+
+	Q_free(buf);
 }
 
-static void SB_Update_Source_From_URL(const source_data *s, server_data *servers[],
-	int *serversn)
+static struct curl_buf *curl_buf_init()
```
fzwoch: For pedantic C make this:
```C
static struct curl_buf *curl_buf_init(void)
```

EX_browser_sources.c
```diff
@@ -197,66 +204,129 @@ static void Precache_Source(source_data *s)
...
+	// Don't modify curl_buf as it might be used to create cache file
+	char *buf = Q_malloc(sizeof(char) * curl_buf->len);
+	memcpy(buf, curl_buf->ptr, sizeof(char) * curl_buf->len);
 
-	while (fgets(buf, sizeof (buf), f) != NULL) {
-		NET_StringToAdr(buf, &addr);
-		servers[(*serversn)++] = Create_Server2(addr);
+	netadr_t addr;
+	// Not using strtok as it's not thread safe
+	for (char *p0 = buf, *p1 = buf; p1 < buf + curl_buf->len; p1++) {
```
fzwoch: For pedantic C make this:
```C
char *p0, *p1;
// Not using strtok as it's not thread safe
for (p0 = buf, p1 = buf; p1 < buf + curl_buf->len; p1++) {
```

krychu:
@fzwoch thanks for your time and CR. I put your suggestions in.

EX_browser_sources.c
```diff
@@ -197,66 +204,135 @@ static void Precache_Source(source_data *s)
...
+		curl_buf_deinit(curl_buf);
+		return NULL;
 	}
 
-	fseek(f, 0, SEEK_SET);
-	
-	SB_Process_URL_Buffer(f, servers, serversn);
-	fclose(f);
-	Q_free(filename);
+	return curl_buf;
```
meag: Add `curl_easy_cleanup(curl);` before returning?

meag:
Will I pull this in and add the `curl_easy_cleanup(curl);` call?

EX_browser_sources.c
krychu: Thanks @meag, it's fixed now.

krychu:
just added that, thanks.

PR #258, opened 2018-04-12
Fix backwards memcpy

tdm4:
In pr_cmds.c line 1774, memcpy aborts on overlapping copies. I changed it to memmove and it no longer SIGABRTs.


meag:
Nice catch - which mod was this causing problems on?

I'll patch mvdsv too, ezquake's server is just a copy of that.

brynet:
There is at least another overlapping memcpy in NQP_Flush:[1774](https://github.com/ezQuake/ezquake-source/blob/7dfe930dd4a7eea7b89a10b9669e527249f7e4e4/pr_cmds.c#L1774) that @tdm4 missed. It should be noted that this is on OpenBSD, where this undefined behaviour triggers a fatal `SIGARBT`.

PR #264, opened 2018-04-29
Fix compiler warnings

javian:
There are a number of compilation warnings when I compile ezquake on macOS High Sierra, this intends to fix the ones related to absolute values as noted in the compiler warnings. I also added some changes to fix some warnings about type conversions.

PR #270, opened 2018-05-07
Fix many abs/fabs/fabsf warnings from the compiler

tdm4:
This PR should fix the warnings generated from using the abs/fabs/fabsf functions.
(This affects clang, I don't think GCC prints these out)


tdm4:
Going to close since #264 is more complete.

PR #274, opened 2018-05-10
Fix typo

drzel:


meag:
Hi @drzel 

I think you want to do a rebase, rather than merging from master?   Also can you rename the commit to "CONFIGS: fix typo in graphics config files", so it's a bit more obvious when reading the commit history.

Thanks,
meag

PR #284, opened 2018-06-24
Some fixes and USE_SYSTEM_MINIZIP build option

evhskurk:
1. "Application controled" => "Application controlled" in help for gl_anisotropy.
2. Add #if guard that prevents endian macros from being redefined.
3. `toggle a b c` command gave incorrect error messages
3b. The regex codepath in Cvar_Toggle_f_base was all over the place, simplified.
4. Added a build option USE_SYSTEM_MINIZIP to allow ignoring the bundled minizip files and use the system's version instead.

Build changes are there because of Gentoo zlib idiosyncrasies requiring me to add `-DOF=_Z_OF` for ezquake to compile with the bundled minizip, and an option to bypass them is nice to have regardless. Does not change the default behavior.


meag:
Closing - this was included in 3.1, github didn't detect.

PR #320, opened 2019-02-12
inbuilt server: QVM from Quake3e, API 16, worked qvm in 64 bit mode, jit compilation, support qvm version 2

angeld29:
* qvm from Quake3e
  worked qvm in 64 bit mode
  jit compilation,
  support qvm file version 2

* Changes in GAME_API_VERSION 16( incompatible with previous)
  server change string pointers in mod memory only in trapcall( strings passed from mod, and placed in mod memory)
  never use pointers to memory outside mod memory, this cannot work in qvm in 64 bit server
  SetString now work only for original dat mods
  mapname and client netnames getting by get_infokey trap

* ******** Required CHANGES in MOD for API 16
  
  * remove all server data from edict_t
    now it must be
    typedef struct shared_edict_s { entvars_t       v;  } edict_t;
  * mod must get client netname in GAME_CLIENT_CONNECT call
    self = PROG_TO_EDICT( g_globalvars.self );
    self->s.v.netname = netnames[NUM_FOR_EDICT(self)-1]; //Init client names
    infokey( self, "netname", self->s.v.netname,  32);
  * mapname can obtain in  GAME_LOADENTS call
    infokey( world, "mapname", mapname, sizeof(mapname) );
  * added parametr to GAME_CLIENT_USERINFO_CHANGED call
    called with 0 before changing, with 1 after changing
    mod must update client netname in call with param 1 and key = "name"


meag:
Hi

Just to let you know you're not being ignored, and to say thankyou very much for working on it.  We'll have to go through a process of getting KTX compatible with any changes though before it can be accepted, so once that's done and we can issue compatible mvdsv & ktx we'll merge this in.

Cheers,
meag

meag:
Hi

Can you describe why the vm*.c files don't match those in ioquake?
Example: https://github.com/ioquake/ioq3/blob/master/code/qcommon/vm_x86.c

Thanks,
meag

angeld29:
1) i take it from Quake3e https://github.com/ec-/Quake3e
2) adapt it to quake1 code( sys_printf, some types)

deurk:
Hey @meag, do you have it in you to deliver this baby to a conclusion (in or out) or is it too much work to make ezq/ktx/mvdsv take this into account?

Thanks!

qqshka:
should be closed, done in mvdsv

tcsabina:
as @qqshka mentioned, this is done elsewhere.
Closing

PR #326, opened 2019-04-18
MACOS: initialize private mouse data pointer

fzwoch:
There are a lot of NULL checks, but the initial value could be anything. "dev" reported some crashes that may be caused by this.


fzwoch:
> If an object that has static storage duration is not initialized explicitly, it is initialized implicitly as if every member that has arithmetic type were assigned 0 and every member that has pointer type were assigned a null pointer constant.

PR #349, opened 2020-04-19
Vsync fix for MacOS

andrestone:
This is the only way to work around this hardcoded limitation on MacOS.
Other cvars have no effect and MacOS offers no GL driver options.


fzwoch:
What does this tackle exactly? For me the current `vid_vsync 0/1` toggle already does what it should?

andrestone:
On my 2020 Macbook Pro, vsync is hard coded in the driver and gets activated after certain thresholds (temperature / clock).

This is a dirty hack that overwrites the instruction coming from the driver.

PR #366, opened 2020-05-21
FRAMEBUFFER: Add `vid_framebuffer_smooth 0` for nearest filtering

Calinou:
This can be used to obtain a crisper appearance when using an upscaled framebuffer.

See <http://tanalin.com/en/articles/integer-scaling/> for more information.

### DIFF
--- a/gl_framebuffer.c
+++ b/gl_framebuffer.c
@@ -35,6 +35,7 @@ extern cvar_t vid_framebuffer_depthformat;
 extern cvar_t vid_framebuffer_palette;
 extern cvar_t vid_framebuffer_hdr;
 extern cvar_t vid_framebuffer_blit;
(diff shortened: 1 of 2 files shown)


hemostx:
Tried this commit. Touching gl_texturemode causes the linear filter to be reapplied to the framebuffer scaling until you vid_restart (you can then proceed to repeat this at will). I had the same problem when I tried to control lightmap filtering with a variable. Sorry, I don't have a fix :(

meag:
Included mostly here: https://github.com/ezQuake/ezquake-source/commit/687e2da23c06fe324f5f4c6a89e3e1770e508953#diff-fd7983ba1d00f010475a65e236499260
Change today to fix toggling on/off

PR #368, opened 2020-05-21
MINOR: Increase the default `cl_physfps_spectator` value to 77

Calinou:
This makes spectator mode look a bit more reactive. The new default value matches most modern QuakeWorld servers' tick rate.

This also makes `cam_thirdperson 1` look significantly less jittery (although some jittering remains on high refresh rate monitors).

This cvar was originally added in eec64e2893a0908ba286a82ba86e536024e524b6 and I understand the rationale for it. That said, I think it's time to revisit the default value in 2020, with high refresh rate monitors being much more common than they were in 2008. Most esports-oriented games don't seem to intentionally lower the physics update rate while spectating; they just apply standard interpolation like we already do (see CS:GO).

If motion sickness while spectating still proves to be an issue, we can add a better way to interpolate POV angles while keeping the high physfps. (A similar method could also be used to fix jittering when using `cam_thirdperson 1`.)


Calinou:
Bump :slightly_smiling_face: Could this be reviewed, please?

tcsabina:
Hello @Calinou ,

Could you tackle the merge conflicts with this PR? I know that it is an old PR, without much attention. But now is the time ;).

Regards,
Toma

PR #373, opened 2020-06-29
USERINFO: include hwid in user's client string

ciscon:


sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
...
+                buffer[fread (buffer, 1, sizeof(buffer) - 1, f)] = '\0';
+                fclose (f);
+        } else if (f = fopen("/etc/hostid", "r")) {
+                buffer[fread (buffer, 1, sizeof(buffer) - 1, f)] = '\0';
+                fclose (f);
+        }
+	memcpy(hwid, buffer, sizeof(buffer));
+
+	return hwid;
+}
```
Calinou, line 767: This snippet has mixed tab and space indentation:

image

sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
 	}
 	return "";
 }
+
+char* Sys_HardwareID(void)
+{
+        FILE *f;
+        char buffer[255] = {0};
+	buffer[0] = '0';
+	char *hwid = malloc(sizeof(char)*255);
```
jite, line 755: sizeof(char) is always 1 and can be removed

sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
...
+
+        if (f = fopen("/etc/machine-id", "r")) {
+                buffer[fread (buffer, 1, sizeof(buffer) - 1, f)] = '\0';
```
jite, line 758: No need to terminate buffer as it's already zero-initialized

sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
...
+                fclose (f);
+        } else if (f = fopen("/etc/hostid", "r")) {
+                buffer[fread (buffer, 1, sizeof(buffer) - 1, f)] = '\0';
```
jite, line 761: same as above

sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
 	}
 	return "";
 }
+
+char* Sys_HardwareID(void)
+{
+        FILE *f;
+        char buffer[255] = {0};
```
jite, line 753: I don't see why you'd need the local buffer at all?

sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
...
+	buffer[0] = '0';
```
jite, line 754: You could do buffer[255] = {'0'}; above and the first entry would be ascii zero while the rest would be zero

cl_main.c
```diff
@@ -1788,7 +1788,9 @@ static void CL_InitLocal (void)
 	Cvar_ForceSet (&cl_cmdline, com_args_original);
 	Cvar_ResetCurrentGroup();
 
-	snprintf(st, sizeof(st), "ezQuake %i", REVISION);
+	char *hwid = Sys_HardwareID();
```
jite, line 1791: Variable declaration should be at top of scope

cl_main.c
```diff
@@ -1788,7 +1788,9 @@ static void CL_InitLocal (void)
...
+
+	snprintf(st, sizeof(st), "ezQuake %i %s", REVISION, hwid);
```
jite, line 1793: free hwid if not used anymore (goes out of scope when this function returns)

sys_posix.c
```diff
@@ -746,3 +746,22 @@ const char* Sys_HomeDirectory(void)
...
+	char *hwid = malloc(sizeof(char)*255);
+
+        if (f = fopen("/etc/machine-id", "r")) {
+                buffer[fread (buffer, 1, sizeof(buffer) - 1, f)] = '\0';
+                fclose (f);
+        } else if (f = fopen("/etc/hostid", "r")) {
+                buffer[fread (buffer, 1, sizeof(buffer) - 1, f)] = '\0';
+                fclose (f);
+        }
+	memcpy(hwid, buffer, sizeof(buffer));
```
jite, line 764: can be removed if local buffer is removed

PR #447, opened 2020-12-01
NEWHUD : skip player inventory items when freefloating

hemostx:
In newhud, going freefloat keeps the last tracked player's inventory on the screen, disconnected from its actual state.
This change makes it behave as per oldhud (and other Quakes etc) where going freefloat does not show the last player's inventory.

### DIFF
--- a/hud_ammo.c
+++ b/hud_ammo.c
@@ -224,7 +224,9 @@ void SCR_HUD_DrawAmmoCurrent(hud_t *hud)
 		proportional = HUD_FindVar(hud, "proportional");
 		always = HUD_FindVar(hud, "show_always");
 	}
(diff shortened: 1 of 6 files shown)


meag:
Hi @hemostx 

Thanks for PR, appreciate it.  Only comment is that in SCR_HUD_DrawBarArmor for example - if in free-floating, this will still call Prepare() whereas the others don't, do we want to make those consistent?  It might affect people if they have other elements linked to the hud elements we're hiding?

Thanks again,
meag

hemostx:
Well, going through the code, I noticed all elements don't necessarily call prepare(), so opted to just leave it as is and watch for side effects. I tested by placing every inventory item independently, and that was fine.
I also traced through prepare() a bit, and didn't fancy seeing all the extra work that it does then getting thrown away, but I guess I need to look at how it's typically used.

PR #464, opened 2021-01-07
INPUT: macOS in_raw, refactor input code

fzwoch:
Simplfy a few parts. Gracefully handle unsuccesful opening of
    the input manager. 'Input Monitoring' permissions must be given
    to the the process in order to use raw input. Also wait for the
    input thread to start/fail before returning main init function.
    That should avoid the potential race condition we had lingering
    there for quite some time.


fzwoch:
I decided to do a little refactoring. I felt like some code was not really needed so did some things I thought would make it look less complex.

There is some added logic on starting the input thread though. It should avoid potential race condition, and should give synchronous feedback of whether opening the raw input was successful or not.

PR #476, opened 2021-01-27
HELP:Update json from code.

hemostx:
Ran dev_help_issues generate for an updated base to work on help strings.
- A bunch of vars/properties got automatically reordered alphabetically.
- A lot of new items to be described have now been system-generated.

### DIFF
--- a/help_macros.json
+++ b/help_macros.json
@@ -1,12 +1,12 @@
 {
   "ammo": {
     "description": "returns amount of ammo held for currently selected weapon",
-    "teamplay-restricted": true,
+    "teamplay-restricted": false,
     "type": "integer"
   },
   "armor": {
     "description": "returns current armor value",
-    "teamplay-restricted": true,
+    "teamplay-restricted": false,
     "type": "integer"
   },
   "armortype": {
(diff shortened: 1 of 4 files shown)


help_macros.json
```diff
@@ -304,25 +304,25 @@
...
     ],
-    "teamplay-restricted": true
+    "teamplay-restricted": false
   },
   "powerups": {
     "flags": [
       "incomplete"
     ],
-    "teamplay-restricted": true
+    "teamplay-restricted": false
```
meag, line 325: Hmmm... this is definitely a bug, but what ruleset did you have enabled when you generated the help file? :)

meag:
Think you've just highlighted a bug :)  Teamplay restrictions are disabled on these macros, but that's just the current state depending on ruleset?

hemostx:
Aha! that was in the default config - default ruleset.
I figured the `generate` command picked from whatever the default is set to, as is suggested by a few vars I checked when I saw what was being removed.

hemostx:
This property does not show up when I `/describe $macro` - only the description does. Maybe that command should actually output a little more for macros?
Hold that thought - this patch showed that /describe does not pull the default values from json but from what the default is currently registered as in memory. The values in there are only useful to an external tool using the json (ie the forthcoming autogenerated reference pages)... so this is where these details will surface for now.

meag:
Pulled in, had to manually fix conflicting files

PR #487, opened 2021-02-07
FEAT: Add easymove from jquake

niclaslindstedt:
Easymove is a mode where you can strafejump simply by holding shift and +forward.

It is usable by people who can't bunny jump properly because of
e.g. a disability or similar, and helps them to be competitive
in games with friends.

It is disabled by default in ruleset smackdown.

The feature was written by Tonik.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,6 +401,7 @@ OBJS_c := \
     cl_cmd.o \
     cl_demo.o \
     cl_nqdemo.o \
+    cl_easymove.o \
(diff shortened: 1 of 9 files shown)


tcsabina:
Hello @niclaslindstedt ,
I am sorry, but I don't see as a good thing.
Bunny hopping/strafe jump is the essence of QuakeWorld. Automatically activating/enabling it is just killing it at all.

PR #521, opened 2021-02-25
MACOS: codesign changes for bundle creation

fzwoch:
ARM builds will require a codesign - period. So we remove the
exisiting signature coming from brew.sh and add some addhoc
signature after install_name_tool is run. It may not pass
Gatekepper, but if one disables Gatekeeper locally you actually
have a chance to correctly run it.

--

With these changes I was actually able to create a M1 bundle that runs.

### DIFF
--- a/misc/install/fixbundle.sh
+++ b/misc/install/fixbundle.sh
@@ -89,7 +89,9 @@ done
 
 function fix_symbols {
 	for DEP in $(get_deps "$1"); do
+		codesign --remove-signature "$1"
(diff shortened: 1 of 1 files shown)


fzwoch:
[ezquake.zip](https://github.com/ezQuake/ezquake-source/files/6044522/ezquake.zip)
Here a snapshot for M1 machines.

Probably you will have to disable Gatekeeper with `spctl --master-disable`

PR #525, opened 2021-03-03
MENU - properly offset cursor image if asked to do so.  change defaul…

ciscon:
properly offset cursor image if asked to do so.  change default value of x offset to 0 to retain old behavior unless offset variable has been set.


meag:
Closing: was included in #528 (?)

ciscon:
yeah, it got wrapped up into that branch, so all is good.

PR #543, opened 2021-04-08
HUD:fix tracker right align w/proportional

hemostx:
Fixes wonky positioning of the killfeed lines when using align_right and proportional fonts.
Computes length from individual glyphs as in VXSCR_MeasureTracker instead of fixed width.

### DIFF
--- a/vx_tracker.c
+++ b/vx_tracker.c
@@ -1209,7 +1209,35 @@ static void VXSCR_DrawTrackerString(float x_pos, float y_pos, float width, int n
 		}
 
 		// Place the tracker.
-		x = x_pos + (align_right ? width - (printable_chars + 1) * width_one_char : width_one_char);
+		if (!proportional) {
+			x = x_pos + (align_right ? width - (printable_chars + 1) * width_one_char : width_one_char);
+		}
+		else {
+			int offset = 0;
+			if (align_right) {
+				for (s = 0; s < trackermsg[i].segments; ++s) {
+					void* pic = trackermsg[i].images[s];
+
+					if (pic) {
+						offset += 8 * 2 * scale;
+					}
+					else {
+						if (trackermsg[i].pad && padded_width && !trackermsg[i].text_flags[s]) {
+							offset += padded_width;
+						}
+						else {
+							offset += Draw_StringLength(trackermsg[i].text[s], -1, scale, proportional);
+						}
+					}
+
+					offset += 8 * scale;
+				}
+				x = x_pos + (width - offset);
+			}
+			else {
+				x = x_pos + width_one_char;
+			}
+		}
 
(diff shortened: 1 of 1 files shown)


meag:
Hi @hemostx 

Can you check if https://github.com/ezQuake/ezquake-source/commit/66c89356fb4c7d9d45649ac2654a443caef3f861 fixes this for you?  Just trying to not measure the strings twice

Thanks

hemostx:
Hey boss, I just rebuilt master with this.
0/1 are now swapped  in terms of where the block of items is on the screen (right 0 = tracker block on the right, right 1 = tracker block on the left)
align_right 0 looks like align_right 1 before the change (block on the right, items don't align to the right of the block)
align_right 1 tracker items align to the right of the block but I guess since the block is on the left of the screen, the width of the right-alignment varies as the items pop in and out
.. and some stuff gets cuts off

(screenshots on the way)

meag:
Hi hemo - ah I think I've fixed r_tracker but hud_tracker is broken?  if no proportional font loaded then r_tracker is still off by one if `r_tracker_proportional` is enabled, but otherwise I think it's behaving as expected - `/r_tracker_align_right` isn't though

hemostx:
Yes, my bad, this is in hud_tracker. I don't remember whether which combinations of proportional and r_ vs. hud_ I tested with. I *think* most if not all of them with the hud_ element.

meag:
Doh, wasn't taking into account x_pos, and my config has `r_tracker_x 0` so didn't spot it
Try again now?

hemostx:
With this commit:

- hud_tracker_align_right 1 : all good
- hud_tracker_align_right 0 : aligns left within the tracker block, but the block remains on the right of the screen.

(same with proportional 0|1)

meag:
Hi hemo

Sorry I think I'm being dense - isn't the hud_ version of the block positioned by the usual hud_editor type variables?  hud_tracker_align_right is just dictating alignment within the block, with r_tracker_align_right the block is always the width of the screen.

Sorry been ages since I looked at this stuff...

Thanks,

hemostx:
You're right. I just checked with hud_editor and it's not a fullscreen width element as shown by the bounding box, the hud_tracker_align_x right in my config made it behave like r_tracker. This patch + alternating align_x left/right makes it all work as a newhud element should behave AFAICT.

Calinou:
@hemostx Off-topic, but what font are you using? It looks really good :slightly_smiling_face:

hemostx:
```
font_facepath                         "ariblk"
font_gradient_alternate_percent       "1"
font_gradient_normal_percent          "1"
font_gradient_number_percent          "1"
font_outline_width                    "4"
```

PR #567, opened 2021-07-18
INPUT: Restore joystick support.

ewhac:
Back in 2013, as part of transitioning to SDL to handle all
platform-specific I/O, support for joystick-style input devices was
removed and not converted over.

This checkin restores joystick-style input support via the SDL APIs.
This is essentially a straight restoration of the original joystick
support mechanism.  While admittedly klunky, it is fully compatible with
the original Quake framework (except as noted below), and any old config
files you happen to have laying around should work without modification.

Changes:
	- Added 'joyindex' cvar, which specifies the index number of the
	  SDL joystick device you want to use.  Original code was
	  hard-coded to device zero; default value of this cvar is zero.
	- Originally, joystick POV/hat switch events were mapped to
	  keycodes AUX[29-31].  Now they have their own keycodes:
	  'JOYPOV{UP,RT,DN,LT}'.
	- Due to a coding bug, the first four joystick buttons were
	  mapped to the keycodes "JOY[1-4]", but everything after that
	  was mapped to "AUX5" and beyond.  Now joystick buttons beyond
	  the first four are mapped to AUX1+, as appears to have been
	  originally intended.
	- The previously implicit limit of 32 joystick buttons is now
	  silently enforced.


tcsabina:
Hello @ewhac ,

I know that it has been a while since this PR was submitted. But what is now the current status of this development?

tcsabina:
Closing as no response from author.

ewhac:
Just a quick note to apologize for not responding to your messages earlier.  Thank you for accepting the change.

PR #581, opened 2021-08-22
HUD: added option hud_armor_hidezero to hide armor number when at 0

ciscon:


ginzberg:
Note: this allows the ezQuake hud to mimic the behavior of the new Quake re-release's default hud.

PR #585, opened 2021-08-28
illumos build fix proposal.

devnexen:
also renaming few types to avoid conflicts with existing ones.

### DIFF
--- a/EX_browser_pathfind.c
+++ b/EX_browser_pathfind.c
@@ -65,12 +65,12 @@ typedef int nodeid_t;
 typedef short dist_t;
 
 // trick around language limits - allows to copy 4 bytes with "="
-typedef struct ipaddr_t {
+typedef struct q_ipaddr_t {
 	byte data[4];
-} ipaddr_t;
+} q_ipaddr_t;
 
 typedef struct ping_node_t {
-	ipaddr_t ipaddr;
+	q_ipaddr_t ipaddr;
 	nodeid_t prev;           // previous node on the shortest path
 	nodeid_t nlist_start;    // index of the first neighbour
 	nodeid_t nlist_end;      // index of the last neighbour + 1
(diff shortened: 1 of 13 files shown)


tcsabina:
Hi @devnexen 
I am not so happy about this, to be honest.
`dir_t` and `file_t` are also used in an other project (mvdsv), so if we want to rename them here, we should do it for consistency in the other project as well.
Does this illumos thingy really brings us dir_t and file_t typedefs? Our linux users don't complain that they cannot compile, so it must be something specific to this specific distro...

devnexen:
yes it s specific to this system.

dsvensson:
Pretty common practice to prefix all project types due to the lack of namespace in C, nothing weird here, other than mvdsv not being a lilbmvdsv.a input to the build?

tcsabina:
> Pretty common practice to prefix all project types due to the lack of namespace in C, nothing weird here...

I am still not convinced that the proposed change worth it.
Currently if you work on the QW API (the protocol between server and client), it is pretty easy to find the link between the mvdsv and ezquake project, because of these shared typedefs.
Renaming them in one project and not in the other, just for the sake of a random linux distribution, is a no-go in my opinion. It just doesn't worth the effort.

I must admit that these typedefs should go into some shared repo instead of them being defined and used in both projects separately, but we are not there yet.

But let this be not my own decision.
@qqshka, if you have some time, could you take a look on this? What do you think?

danielstravito:
It's not Linux, it's a Solaris derivative, before you know it there's Haiku and Tru64 PR's all over the place :)

dsvensson:
It's not Linux, it's a Solaris derivative, before you know it there's Haiku and Tru64 PR's all over the place :)

tcsabina:
after an internal discussion we have decided not to merge this PR.

PR #589, opened 2021-09-06
OSX: fix qw url handling

ciscon:
qw urls are now passed in to ezquake from osx via drag and drop emulation, so handle that use case and do not attempt to open qw:// urls as demos.  init_url_handler called from sys_posix.c and defined in sys_osx.m probably aren't needed anymore, leaving in case there is still some use case for that logic.

### DIFF
--- a/common_draw.c
+++ b/common_draw.c
@@ -28,6 +28,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "gl_model.h"
 
+void Draw_ClearConback(void);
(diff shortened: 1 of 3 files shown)


ginzberg:
Wanted to bump this PR. Having proper qwurl behavior is a big QoL boost.

meag:
Should line 906 be deleted in the SDL_DROPFILE handling?

ciscon:
not sure i understand what you're getting at

PR #635, opened 2022-02-14
BINDS: default weapon next/prev binds using weapon command instead of…

ciscon:
… impulse which interferes with weapon preselection.  changing to impulse.

### DIFF
--- a/menu_options.c
+++ b/menu_options.c
@@ -1126,8 +1126,8 @@ setting settbinds_arr[] = {
 	ADDSET_BASIC_SECTION(),
 
 	ADDSET_SEPARATOR("Weapons"),
-	ADDSET_BIND("Previous Weapon", "weapon 12"),
-	ADDSET_BIND("Next Weapon", "weapon 10"),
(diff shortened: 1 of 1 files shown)


meag:
Hi ciscon

What script combination is this fixing?  It breaks my weaponscript for example, and the change from impulse to weapon was made because people were complaining in the other direction.

meag

tcsabina:
@ciscon ,

What's up with this one?

ciscon:
this was an old thing someone reported, using prev/next weapon binds calls weapon X instead of impulse, thereby replacing the previous weapon preselection priority list instead of just affecting what the next +attack would be.  if @meag had a reason to switch it to this then it's probably going to upset someone one way or the other.

tcsabina:
> this was an old thing someone reported, using prev/next weapon binds calls weapon X instead of impulse, thereby replacing the previous weapon preselection priority list instead of just affecting what the next +attack would be. if @meag had a reason to switch it to this then it's probably going to upset someone one way or the other.

So leave it as it is? Meaning to close the PR without merge?

PR #636, opened 2022-02-14
BUILD: support netbsd

ciscon:


q_shared.h
```diff
@@ -156,7 +156,7 @@ float	FloatSwapPDP2Lit (float f);
 #endif
 
 //======================= FreeBSD/OpenBSD DEFINES ====================================
```
VVD, line 158: Add NetBSD to this comment.

EgoPenguin:
libjpeg-turbo conflicts with jpeg-9d a pre-installed package. It works without it. Else no errors or anything while compiling.

q_shared.h
ciscon: all set

PR #638, opened 2022-02-21
OSX: qw urls are now passed in to ezquake from osx via drag and drop …

ciscon:
OSX: qw urls are now passed in to ezquake from osx via drag and drop emulation, so handle that use case and do not attempt to open qw:// urls as demos.  init_url_handler called from sys_posix.c and defined in sys_osx.m probably aren't needed anymore, leaving in case there is still some use case for that logic.


ciscon:
@meag just redid it so it'd be cleaner

PR #649, opened 2022-04-16
SBAR: Add scr_scoreboard_showflagstats 0|1.

dsvensson:
Previously these columns were only available for Team Fortress, but they're valuable for CTF game mode as well.


dsvensson:
Typically enabled via `alias on_enter_ctf "scr_scoreboard_showflagstats 1"`.

PR #652, opened 2022-04-20
SBAR: Avoid GA overflow when carrying regeneration rune.

### LINKED ISSUE #651: hud_bar_armor overflows with regeneration rune in CTF
When carrying the regeneration rune in CTF all armors can regenerate up to 150 points. This works fine for YA where 150 is max, and RA where max is more, but for green armor the armor bar grows outside its intended drawing region.

https://github.com/ezQuake/ezquake-source/blob/6c0f284375c66d9733943cf367585fa83f31b09e/hud_armor.c#L210-L213

Maybe it could work similar to how mega works for health?

https://github.com/ezQuake/ezquake-source/blob/6c0f284375c66d9733943cf367585fa83f31b09e/hud_health.c#L103-L106

image

...not sure about the name though... `color_ga_regen` sounds a bit too specific, perhaps there are other mods that can go above 100 for GA? `color_ga_(plus|over|above|overflow)`?

dsvensson:
Fixes #651

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -6787,6 +6787,10 @@
       "group-id": "19",
       "type": ""
     },
+    "hud_bar_armor_color_ga_over": {
+      "group-id": "19",
+      "type": ""
+    },
     "hud_bar_armor_color_noarmor": {
       "group-id": "19",
       "type": ""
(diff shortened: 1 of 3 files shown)


dsvensson:
Sample:
image

PR #661, opened 2022-07-07
updated build-linux.sh with arch support

ginzberg:
I created a brand new arch VM and piece-wise added packages until I was able to successfully build ezquake. I believe that this should work for others. Note - arch is a rolling release distribution, so I have omitted the version check. 
Also - I updated the pacman command to skip already installed packages, and to skip user intervention approve the packages to be installed.


ginzberg:
One thing that i'm unsure of is how this may work for arch-derivatives, or what would need to be updated to support them - for example, Manjaro. If anyone is more familiar with that, would love to hear it.

eGax:
I'm on Manjaro Linux. Do you have a PKGBUILD for this yet on the AUR?

ginzberg:
@eGax there is `ezquake` and `ezquake-git` on the AUR, but this PR is unrelated. This is to update the build script that is within the main ezQuake github repo. It looks like `ezquake` AUR is still ezQuake 3.2, but I presume that `ezquake-git` would be 3.6. 

After my last comment I realized that supporting Manjaro in the `build_linux.sh` script may really just come down to including the ID from `/etc/os-release` in the case block, and have it simply replicate the same logic as the `arch` case. I don't have a Manjaro install handy to test that however.

PR #662, opened 2022-07-07
removed commands that throw errors from all gfx_gl_*.cfg

ginzberg:
I think that there's a separate opportunity to modernize these settings as there's been a lot of new options to our renderer (outlines, drawflat filters, etc) but I wanted to first eliminate the commands that no longer are relevant from our cfgs.

One note - there's really no great way to distribute these files. It really depends on these updates making their way into the ezquake and nquake packages. That said, still feel it's valuable to make the relevant updates.


mushis:
I haven't checked the changes, but kudos for the cleanup. it was needed.

PR #668, opened 2022-07-27
SMARTSPAWN: Do not fire with +attack to spawn

JosephPecoraro:
**Adds:** `cl_smartspawn`
  - `0` disabled (default)
  - `1` enabled

**Description:**
When dead, `+attack` will spawn and immediately `-attack` to prevent weapon firing when spawning.

**Rationale:**
A best practice is to spawn without shooting. Spawning by shooting has numerous disadvantages:
  * wastes ammo
  * prevents an aimed shot for at least one reload cycle
  * can accidentally hurt or kill a teammate (Murphy's Law)
  * may reveal your spawn location via bullet impacts on a far wall to an enemy that otherwise
    would not know
  * inadvertently distorts a players true sg efficiency as most spawn shots are misses

**Solution:**
Add a command, like `cl_smartjump`, that better matches the player's intended behavior.

**Implementation:**
When a player is dead, an attack command (`+attack` / `+fire`) is treated as a jump instead
of an attack.

**Rulesets:**
Allowed by all. This is easily achievable with weapon scripts.

### DIFF
--- a/cl_input.c
+++ b/cl_input.c
@@ -31,6 +31,7 @@ cvar_t cl_c2sImpulseBackup = { "cl_c2sImpulseBackup","3" };
 cvar_t cl_c2sdupe = { "cl_c2sdupe", "0" };
 cvar_t cl_forwardspeed = { "cl_forwardspeed","400" };
 cvar_t cl_smartjump = { "cl_smartjump", "1" };
+cvar_t cl_smartspawn = { "cl_smartspawn", "0" };
 cvar_t cl_iDrive = { "cl_iDrive", "0", 0, Rulesets_OnChange_cl_iDrive };
 cvar_t cl_movespeedkey = { "cl_movespeedkey","2.0" };
 cvar_t cl_nodelta = { "cl_nodelta","0" };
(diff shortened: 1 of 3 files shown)


JosephPecoraro:
50% of why I did this is so that dracs will stop messaging me about it.

JosephPecoraro:
Might need an update to `release-notes-latest.md`.

teiman:
This might be a wellcome change to some people. But is somewhat hardcoding a behavior in the engine that is really a gamelogic issue in progs.dat
Is the programming in QuakeC by the mod that define what means to be dead and revive. This programming usually add a cooldown after the revive, so you can't shot immediately. If this fails, is a bug in the mod, not the engine.

Anyway since is optional and off by default, I don't see any problem with this other than feature creep.

tcsabina:
What is the purpose of this? When you die, just hit the jump instead of fire if you don't want to shoot. Simple... 
Why would we need this at all?
I want a fire button to fire. Always. Not a modified behavior based on some conditions

zigg1zagg1:
If the implementation of this is the same as in unezQuake, then it needs some tweaking, there it works for +attack but +fire will still cause you to shoot when spawning

jite:
Not using +attack to spawn has always been part of the skillset and the gameplay. Some games you Want to use it to try and get the last frag in an even game… Sometimes you kill and enemy, sometimes you fkup and kill a teammate. It’s part of the game.

tcsabina:
alright!
Last nail in the coffin for this PR was the comment from dimman. Closing.

Issue #665, opened 2022-07-19
FEATURE: Deal with the low memory limits.


dsvensson:
Modern maps may need to start ezquake with -mem 256 rather than the default. While this might seem trivial I spent an hour helping out two players with adding that command line parameter via voice. One Windows and one MacOS user. It may sound odd that people don't know what command line parameters are, but that's just the way it is apparently.

The simple solution would be to just increase the limit like other engines already have. Another solution might perhaps be some 2x resize or some other resize strategy if the map hits the limit until it loads, and let the OS oom-killer slay the process if it gets out of hand.

If I'm interpreting the FTE source code correctly it looks like -mem was removed in 2013 in favor of dynamic allocation as needed.

Calinou:
Defaulting to `-mem 256` should work out fine on machines with 4 GB of RAM or more. This could be a reasonable solution for now.

inf1niti:
IME, it's still possible to get crashes even with a higher heapsize allocation - but this has only happened to me while loading multiple large maps, and when warping between them throughout the course of a night of games.  So there may also be some issues around bspfile cleanup between map loads.  

It's also worth noting that any memory increases would **need to be implemented on the MVDSV side as well**.

Beyond global memory allocation, there are a number of issues IME with unexpectedly small limits, or strange bugs when trying to render more complex maps:  

- [ ] [textures larger than 640*480 throw a `sys_error`](https://github.com/ezQuake/ezquake-source/blob/0faefe4544f895f4c4686fe768dda8129a94d7b4/r_texture_load.c#L559) (many common SP wads these days contain 1024x1024 textures).

- [ ] arbitrary/hard-coded limits are present in [bspfile.h](https://github.com/ezQuake/ezquake-source/blob/6c0f284375c66d9733943cf367585fa83f31b09e/bspfile.h#L26-L45), [client.h](https://github.com/ezQuake/ezquake-source/blob/6c0f284375c66d9733943cf367585fa83f31b09e/client.h#L306-L307), and [common.h](https://github.com/ezQuake/ezquake-source/blob/6c0f284375c66d9733943cf367585fa83f31b09e/common.h#L71-L77) headers which cause various load problems if any of these limits are exceeded (even if they fall within the limits imposed by the bsp29 format, and are properly rendered in other engines). 

[...]

PR #674, opened 2022-09-02
LIMITS: Bump limits to allow loading modern maps.

dsvensson:
This fixes various parts of #665 and loads the new rctf maps correctly.


dsvensson:
The final component to load modern maps is 1024x1024 textures support which is found here:

https://github.com/ezQuake/ezquake-source/commit/869bc51e847a946e0c44ade61350f3f83ee61743

PR #681, opened 2022-09-15
HUD: Show flag carrier in hud_teaminfo.

dsvensson:


tcsabina:
Nano,

Can we see a before/after screenshot ;).

Thanks

hud_teaminfo.c
```diff
@@ -629,10 +666,26 @@ void Parse_TeamInfo(char *s)
...
 	ti_clients[client].armor = atoi(Cmd_Argv(6));
-	ti_clients[client].items = atoi(Cmd_Argv(7));
+	ti_clients[client].items = Filter_FlagsAndRunes(client, atoi(Cmd_Argv(7)));
 	strlcpy(ti_clients[client].nick, Cmd_Argv(8), TEAMINFO_NICKLEN); // nick is optional
 }
 
+void Update_FlagStatus(int player_num, char *team, qbool got_flag)
+{
+	int flag = IT_KEY1 | IT_KEY2;
+	if (!strcmp(team, "blue")) {
```
dsvensson, line 676: Need to revisit this tonight, might have mixed up the teams vs it_key's... in practice it doesn't matter as it's reversed in both mapping and UI condition as the correct flag char/icon is shown, but would be nice if the code makes complete sense. :)

dsvensson:
image

correct mapping:

https://github.com/QW-Group/ktx/blob/3e7da9402976b015f18b7e236e942f5a5bb3a139/src/ctf.c#L283-L288

PR #692, opened 2022-10-08
Follow up of PR691

tcsabina:


VVD:
Look like use before initialization.

tcsabina:
@ciscon @dsvensson , how does this PR look like? Does it tackle all the issues from PR 691?

@VVD , where exactly?

dsvensson:
Particle trails no longer render when fog is enabled, haven't checked other particle effects so maybe more affected. Haven't had time to check why, and it's slightly outside of my skill set. Maybe it's just some shader ordering or something. At the same time it would suck if the fog stuff bitrotted unmerged as it looks great otherwise.

dsvensson:
The texture commit can be merged separately if you put this on hold, it works great.

PR #695, opened 2022-10-24
CI: Add macOS build targets.

dsvensson:


dsvensson:
Can't be unified in the same runner via strategy as they both install Homebrew packages and would then overwrite each other. Verbose, but works. Still waiting for someone to properly test the build artifacts.

Makefile
```diff
@@ -65,10 +65,19 @@ else
...
+        LDFLAGS_c += -headerpad_max_install_names
+
+        # Cross-compiling support
+        ifneq ($(DARWIN_TARGET),)
+            CFLAGS_c += -target $(DARWIN_TARGET)
+            LDFLAGS_c += -target $(DARWIN_TARGET)
+        else
+            CFLAGS_c += -mmacosx-version-min=10.8
+            LDFLAGS_c += -mmacosx-version-min=10.8
+        endif
```
dsvensson, line 80: Would be great if someone who enjoys macOS development untangles this, but it works at least.

dsvensson:
arm64 build verified by dev to work just as well as a local build.

PR #700, opened 2022-11-13
ALLOWMULTIPLE: don't restrict to debug builds

dusty-qw:


tcsabina:
@dusty-qw ,
should we remove the -allowmultiple command line option as well? I see no point keeping it as requirements. Just like the linux version, there is no command line option needed to spawn as many clients as you want.

PR #702, opened 2022-11-13
SMACKDOWN: allow gl_outline and particle/custom LG

dusty-qw:
Hopefully this isn't too controversial. These changes have been a long time coming: gl_outline was excluded due to a bug that has now been fixed for almost two years, and custom LGs models are a non-issue now that everyone uses fakeshaft anyway. These are both popular features and many would appreciate the update.

### DIFF
--- a/help_cmdline_params.json
+++ b/help_cmdline_params.json
@@ -1,8 +1,5 @@
 {
   "-allowmultiple": {
-    "builds": [
-      "debug"
-    ],
     "description": "On Windows, launch multiple copies of ezQuake rather than re-using the existing instance",
     "systems": [
       "windows"
--- a/rulesets.c
+++ b/rulesets.c
@@ -96,8 +96,7 @@ qbool RuleSets_DisallowModelOutline(struct model_s *mod)
 		case MOD_BACKPACK:
 			return !cls.demoplayback && (rulesetDef.ruleset == rs_qcon || rulesetDef.ruleset == rs_smackdown);
 		default:
-			// return to just rs_qcon once backface outlining tested
-			return !cls.demoplayback && (rulesetDef.ruleset == rs_qcon || rulesetDef.ruleset == rs_smackdown);
+			return !cls.demoplayback && (rulesetDef.ruleset == rs_qcon);
 	}
 }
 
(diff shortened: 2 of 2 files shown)


tcsabina:
Hi Dusty!

I think we should at least hear the opinion of the tournament admins on this. Let me try to come up with the conversation on discord.

tcsabina:
@dusty-qw ,
The `gl_outline` stuff is straight forward, but what other features gets enabled with this in Smackdown? `r_shaftalpha`?

dusty-qw:
> @dusty-qw , The `gl_outline` stuff is straight forward, but what other features gets enabled with this in Smackdown? `r_shaftalpha`?

Yes. Particle LG allows you to specify color and alpha as well as using the fancy particle "bolts". Keep in mind smackdown already allows you to make the LG bolt "invisible" by setting its color to be same as your wall/floor colors.

zigg1zagg1:
makes sense to also allow r_shaftalpha if you allow particle shaft + custom shaft model
i think this might be one of those things a lot of players dont have an issue with but dont use because its disallowed,
i know a few ppl used shaftalpha back when thunderdome ruleset was allowed in 4on4 tourneys

hemostx:
This could maybe just become a fork of smackdown (finding a new "common" name is left as an exercise to the reader), that way qcon and smackdown can remain as fixed as i.e. the Reference Kilogram as can be for posterity/comparison/default "safe" rulesets when other options can't be agreed upon.

zigg1zagg1:
Yeah not a bad idea, currently theres no way to allow shaftalpha but still disallow things like fullbright backpacks.
Tho i guess it might be less likely for players to switch if its on a different ruleset, its up to tournament admins whats allowed in the end

dusty-qw:
> makes sense to also allow r_shaftalpha if you allow particle shaft + custom shaft model i think this might be one of those things a lot of players dont have an issue with but dont use because its disallowed, i know a few ppl used shaftalpha back when thunderdome ruleset was allowed in 4on4 tourneys

The shaftalpha check is tied to the particle restrictions, so allowing particle LG also allows r_shaftalpha.

qwhammer:
Makes absolutely no sense to allow those cheats on smackdown ruleset.
To whom wants to use it should then use thunderdome ruleset or default.
Not sure about other event admins but on HTe will definitely not be allowed.

tcsabina:
closing, as agreed with Dusty to handle these separately.

PR #709, opened 2022-11-29
VID: enable window notification upon activity in linux/macos

ciscon:


ciscon:
until someone builds a proper statically linked sdl2 with the newer release we're sol on this

Issue #630, opened 2022-01-28
REQ: Support for alpha-channel map entities in EZQuake


inf1niti:
**Is your feature request related to a problem? Please describe.**
Other modern Quake engines support alpha transparency on `func_wall` and `func_illusionary` entities in a bsp file.  this is supported via the `alpha` property on brush entities.  

This effectively lets a map implement semi-transparent brushes that can be see-through, like vis'd water, to represent windows, lasers, glass doors, etc.  

**Describe the solution you'd like**
I'd like ezquake to be able to support this so it can align better with where other quake engines are at today.  Although ezquake is more gamplay focused, this is a change that effects gameplay as much as aesthetics because transparent brushes effect gameplay. 

**Additional context**
IIRC support for this was added in protocol changes to the original quake protocol.  As such, support might need to be added through updating the QW protocol as well.

**Impact on server or mods**
If there are necessary protocol updates, this would have an impact on (or would require updates to) KTX/mvdsv as well.

meag:
Can you supply an example .bsp file and qwd/mvd file with the alpha property being sent?

inf1niti:
So I'm not actually sure the alpha property gets sent - because its `func_wall` or `func_illusionary` it is non-vis blocking, so you can see things through it.  As long as it's being rendered appropriately in the client it should be fine?

In standard quake however, support for this feature was added in Fitzquake and rolled under a protocol change (protocol 666), and it doesn't work in protocol 15 network play (that's why I mentioned it).  **Spike** could likely provide more details if necessary (this works in multiplayer using FTE as a server).

## Example 

Here's an [example BSP](https://drive.google.com/file/d/12-rTHOCemGlKDi12AdoFdTpfL_l9A-To) using the alpha property on a window element.  

| Details | Screenshot |
| --------- | ----------- |
| setting property in trenchbroom | ![image](https://user-images.githubusercontent.com/67801374/151649542-d82c457b-712e-4a85-ae7e-99bd32184ba3.png) |
| Rendered in EZQuake |  ![ezquake035](https://user-images.githubusercontent.com/67801374/151649694-7953e175-2b16-4d8b-b3ad-585fb4b66bec.png) |
| Rendered in QSS | ![qssm0495](https://user-images.githubusercontent.com/67801374/151649717-5f87e17f-1409-4aff-9939-61049821f25f.jpg) |

dsvensson:
[...]

PR #714, opened 2022-12-14
NET: Support FTE transparency extension.

dsvensson:
This is a straight port from the FTE client implementation on top of the support that was already in ezQuake.

https://github.com/fte-team/fteqw/blob/16c8e521ef8fe1a8d47dcaa36207cd5eb0f09dce/engine/common/protocol.h#L600-L601

https://github.com/fte-team/fteqw/blob/16c8e521ef8fe1a8d47dcaa36207cd5eb0f09dce/engine/client/cl_ents.c#L4929-L4939

This fixes the network part of #630

### DIFF
--- a/cl_ents.c
+++ b/cl_ents.c
@@ -1366,7 +1366,19 @@ void CL_ParsePlayerinfo (void)
 	} 
 	else 
 	{
-		flags = state->flags = MSG_ReadShort ();
+		flags = (unsigned short) MSG_ReadShort ();
+
+		if (cls.fteprotocolextensions & FTE_PEXT_TRANS)
+		{
+			if (flags & PF_EXTRA_PFS)
+				flags |= MSG_ReadByte() << 16;
+		}
+		else
+		{
+			flags = (flags & 0x3fff) | ((flags & 0xc000) << 8);
+		}
+
+		state->flags = flags;
 
 		state->messagenum = cl.parsecount;
 		if (cls.mvdprotocolextensions1 & MVD_PEXT1_FLOATCOORDS) {
--- a/protocol.h
+++ b/protocol.h
@@ -285,8 +285,8 @@ void MSG_DecodeMVDSVWeaponFlags(int flags, int* weaponmode, int* weaponhide, qbo
 // bits 11..13 are player move type bits (ZQuake extension)
 #define PF_PMC_SHIFT	11
 #define	PF_PMC_MASK		7
-#define	PF_ONGROUND		(1<<14)			// ZQuake extension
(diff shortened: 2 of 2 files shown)


dsvensson:
Demo of ezquake connected to FTE server running NQ CRMod+

ezquake000

tcsabina:
Hi nano!

You know that protocol.h is used in mvdsv, and mvdsv parser as well, right?

dsvensson:
Humm... right.. didn't test this with the built-in mvdsv, only external. Going to try that as well. The mvd section is separate though.

tcsabina:
> Humm... right.. didn't test this with the built-in mvdsv, only external (where the non-trans else case is used). Going to try that as well. The mvd section is separate though.

Well, I am not really worried about the built-in mvdsv. I have some concerns about modifying the protocol.h file, which is the "QW API", the link between server-mod-client. This file has the version (currently 28), which at least should be modified/updated with a change.
But I am not sure what exact measures must be taken when we want to update this file.

dsvensson:
As you see in that file it already supports a number of FTE extensions. The change of the PF_SOLID and PF_ONGROUND are kind of virtual. If PF_EXTRA_PFS (**P**layer**F**lag**S**) is set, an extra byte is read and then those two are read at the new offset. If not, then the bits are reshuffled in the incoming data from 14 and 15 to the 22 and 23. That way the code that deals with PF_SOLID/PF_ONGROUND doesn't have to consider PF_EXTRA_FPS to know where those two flags are to be read. Can probably be implemented differently, just went with the same implementation as FTE. Makes sense to add a comment to it as it took a while to get when reading the FTE code. The embedded mvdsv needs to account for this as well, while the external one remains unchanged until it's time to add support for PF_EXTRA_PFS to it as well to fully support alpha vs the current aliasmodel limited alpha-support mvdsv/ktx has now for race support.

cl_ents.c
```diff
@@ -1366,7 +1366,19 @@ void CL_ParsePlayerinfo (void)
...
+		flags = (unsigned short) MSG_ReadShort ();
+
+		if (cls.fteprotocolextensions & FTE_PEXT_TRANS)
+		{
+			if (flags & PF_EXTRA_PFS)
+				flags |= MSG_ReadByte() << 16;
+		}
+		else
+		{
+			flags = (flags & 0x3fff) | ((flags & 0xc000) << 8);
```
dsvensson, line 1378: :construction: :warning:  The embedded mvdsv must account for this before merge. :construction: :warning:

dsvensson:
Closing this for now to avoid confusion. Opening a new PR once mvdsv support has landed.

PR #716, opened 2022-12-15
BSP, MODERN: Add support for lit water.

### LINKED ISSUE #631: REQ: support for lit water (`-splitturb` in QBSP) [mapping]
**Is your feature request related to a problem? Please describe.**
Other quake engines have recently moved to add support for `-splitturb` in QBSP, which allows for lighting in the map to effect water textures so they're not rendered as fullbrights.  This provides more natural looking lighting and enhances the visuals of maps.  

**Describe the solution you'd like**
Would be nice for EZQuake to support this feature and align with the rendering support that's happening in other engines.

**Additional context**
[implementation in QSS](https://github.com/Shpoike/Quakespasm/commit/4ec4cde5f4f157681de7d00ae2deb7c787bdd093)
**comparative screenshots:**
image
image
image

**Impact on server or mods**
none.

dsvensson:
fixes #631

### DIFF
--- a/gl_model.h
+++ b/gl_model.h
@@ -67,6 +67,7 @@ typedef struct texture_s {
 	qbool               loaded;                     //help speed up vid_restart, actual only for brush models
 	int                 isLumaTexture;              //fb is luma texture, rather than normal fb
 	qbool               isAlphaTested;              //texture is flagged for alpha-testing
+	qbool               isLitTurb;                  //turb texture has lightmap
 	int                 turbType;
 
 	int                 gl_width;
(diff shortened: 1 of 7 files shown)


mushis:
1.  this requires different maps, correct? (different.bsp files)
2. if I have a version of dm3 with lit water, it won't pass the f_modified check, right?

Just want to make sure people can't use this to cheat.


bonus q: this is also know as vis/vised map, right?

dsvensson:
1. It requires maps compiled with -splitspecial/-splitturb, see ericw-tools for details.
2. The hashes involved provide no real cheat prevention as this is an open source project. You don't know if a client is modified as there's no chain of trust. But you're right that an unmodified client with a recompiled map would not pass hash checks, would be cool to recompile all old maps with modern tooling. Can however be solved by mainstreaming the use of version-suffix of maps, ie `e2m2_v2.bsp` etc.That way maps never have to be considered finished, but can evolve over time as better tooling, and better design ideas pop up. 

bonus a: No, it's known as lit water or lit turb, all mappers vis their maps to simplify the computations of the light utility, not all pass -splitspecial/-splitturb however. Maybe you're thinking about the old vispatch.exe utility. This PR is just about lightmaps on turb surfaces.

tcsabina (approved):

PR #717, opened 2022-12-18
Allow gl_outline in smackdown

tcsabina:
Allow `/gl_outline` in smackdown ruleset, just like `/r_fx_geometry`.
Disallow both in qcon ruleset.

2 new entry in the Options for these two outline settings

### DIFF
--- a/rulesets.c
+++ b/rulesets.c
@@ -81,6 +81,7 @@ qbool RuleSets_DisallowSimpleTexture(model_t* mod)
 	}
 }
 
+// for models (gl_outline)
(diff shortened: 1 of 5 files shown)


rulesets.c
```diff
@@ -90,14 +91,25 @@ qbool RuleSets_DisallowModelOutline(struct model_s *mod)
 
 	switch (mod->modhint) {
 		case MOD_EYES:
-			return true;
```
dusty-qw, line 93: You want eyes to have outline?

PR #721, opened 2022-12-22
PIPELINE: add python-tk dep

ciscon:
testing to see if this fixes macos intel builds


ciscon:
brew is just a broken mess and that will "hopefully" get fixed sometime in the near future, closing this out.

PR #722, opened 2022-12-22
MODERN: Fix alphatest related fog/fence issues.

dsvensson:


dsvensson:
Changing the threshold:
alpha-variants-fs8

Effect of setting alpha after discard, within alphatest:
alpha-steep-fs8

More on steep angles where the shader changes produce transparent texture instead of black texture on steep angles:
alphafog-fs8

Issue #671, opened 2022-09-01
BUG: remastered looping sounds don't loop since broken looping sound fix


hemostx:
**ezQuake version:**
starting from 95f4f98203be115994ef82fe12c8eb1cd6a31a5e

**OS/device including version:**
W10 21H2

**Describe the bug**
Since merging master (which includes 95f4f98203be115994ef82fe12c8eb1cd6a31a5e and up), the remastered sounds I extracted from the steam rerelease into a pk3 will not loop. Anything one-shot plays just fine, otherwise it's `Sound ambience/whatever not looped`.
- Rebuilt only master to confirm it's not my branch
- Go back one commit before that one: these sounds loop again
- removed `SDL_AUDIODRIVER directsound` env variable to switch to wasapi and exclude sound API interference: same behaviour on either.
- I wrote "from a pk3" because somehow I can not get the flat files into id1/sounds or w/e in a way that my client wants to play any of them (looping or not!). I will rule that out as heat-related head trauma on my end, and not a pk3 vs flat file read issue

**To Reproduce**
1. Setup a pk3 with some ambient sounds from last year's remaster (I can lend you mine if you need one), put in id1 or qw
2. Default config, but make sure s_ambientlevel is audible
3. map start
4. Console is spammy, game world is quiet.

**Expected behavior**
Looping loops :)

hemostx:
I also played around with s_khz and s_linearresample, with no effect.

PR #723, opened 2022-12-23
SOUND: when loading an external ambient sound a lockup occurs

ciscon:
SOUND: when loading an external ambient sound (eg water1.wav), in S_ParseCueMark where it calls BuffLittleLong, it always returns the size as 0, causing it to loop forever

### DIFF
--- a/snd_mem.c
+++ b/snd_mem.c
@@ -669,10 +669,11 @@ sf_count_t SFVIO_Tell(void *user_data)
 static qbool S_ParseCueMark(const byte* chunk, int len, int cue_point_id, int* sample_length)
 {
 	int pos = 0;
+	unsigned int size = 1;
 
 	*sample_length = 0;
-	while (pos < len - 8) {
-		unsigned int size = BuffLittleLong(chunk + pos + 4);
+	while ((pos < (len - 8)) && (size != 0)) {
+		size = BuffLittleLong(chunk + pos + 4);
 
 		// Looking for ltxt chunk with purpose "mark"
 		if (size >= 20 && !strncmp(chunk + pos, "ltxt", 4) && !strncmp(chunk + pos + 16, "mark", 4)) {


ciscon:
it leaves ambient sounds broken with external sounds, so there's definitely a break in the way they are being handled in general, but at least it doesn't lock up.  using the old wav loading works.

dsvensson:
Sweet, I actually hit this as well when running crmod on ftesv and connecting with ezq. The map in question uses `ambient_general` that resulted in ezquake freezing on connect. This PR fixes that issue.

hemostx:
Ooh, I wonder how that plays with #671 (remastered ambient loops don't play anymore) which is a result of 95f4f98203be115994ef82fe12c8eb1cd6a31a5e - (Fixed playback of looping sounds when using libsndfile to load sounds (reported by lurq)).

I can only check if this impacts #671, as I did not encounter the issue fixed by 95f4f98203be115994ef82fe12c8eb1cd6a31a5e.

snd_mem.c
```diff
@@ -669,10 +669,11 @@ sf_count_t SFVIO_Tell(void *user_data)
 static qbool S_ParseCueMark(const byte* chunk, int len, int cue_point_id, int* sample_length)
 {
 	int pos = 0;
+	unsigned int size = 1;
 
 	*sample_length = 0;
-	while (pos < len - 8) {
-		unsigned int size = BuffLittleLong(chunk + pos + 4);
+	while (pos < len - 8 && size !=0 ) {
```
tcsabina, line 675: Can you add some braces around the logical operations, please?

Also, there should be a space before 0, and nospace before the closing brace ;)

tcsabina (changes requested):

tcsabina:
@ciscon ,
Can you add those cosmetic changes?

tcsabina (approved):

PR #727, opened 2023-01-05
MINOR: Warn on empty textures rather than break.

dsvensson:
This is a common annoyance when TrenchBroom inserts a __TB_Empty texture due to some action when mapping. The result is that the size check later on causes a division by zero which crashes ezQuake without giving any actionable feedback to the user.


dsvensson:
lol... not only ezq having this exact issue, with the exact same fix :)

https://github.com/Novum/vkQuake/issues/567#issuecomment-1264342014

PR #728, opened 2023-01-08
Using the qwprot repository as submodule

tcsabina:
Started to use the qwprot repository. Removed protocol.h as it is part of qwprot.


dsvensson:
Here I mean, submodules:true to checkout step in the CI yml files.

VVD:
Don't forget to add all submodules to archive with sources.

PR #731, opened 2023-01-12
BUILD: update build-linux.sh script to checkout submodules

ciscon:


ciscon:
fixed macos targets as well

PR #742, opened 2023-01-27
increase fps when disconnected

zigg1zagg1:
Using the mouse at 30fps on the menu feels bad, this changes it to use the refresh rate instead.

Tested and it works for me but would appreciate if it does for others i barely understand what im doing :P


src/cl_main.c
```diff
@@ -2114,7 +2114,7 @@ static double CL_MinFrameTime (void)
 		return 0;
 
 	if (cls.state == ca_disconnected || (Minimized && !cls.download))
-		return 1 / 30.0;
+		return 1 / r_displayRefresh.value;
```
qqshka, line 2117: Or you could remove `cls.state == ca_disconnected` and that would be better since with the change you did you also affected FPS for a minimized situation and it would be better to not hog CPU in such a situation.

tcsabina: so @zigg1zagg1 , are you going to update this PR based on qqshka's reply?

zigg1zagg1:
Should be ok now? 
didnt mean to add the protocol file, only had that in for testing
Still figuring this out :|

tcsabina:
> Should be ok now? didnt mean to add the protocol file, only had that in for testing Still figuring this out :|

I think it is the other way around ;).
`(Minimized && !cls.download)` should stay and `cls.state == ca_disconnected` should be removed instead.


quoting qqshka:
> Or you could remove `cls.state == ca_disconnected` and that would be better since...


@qqshka , right? That is what you've meant.

qqshka:
Yeah, you should just remove `cls.state == ca_disconnected` in the original code. But that should be done in ONE commit, not 4 since git would keep that in history and it is ugly. So remove all previous commits and do it in one and then force push the branch.

And I still think it could have a negative impact on people who use 'cl_maxfps 0', they probably would have thousands of FPS when disconnected which could result in extreme heat and coil whine, oh well.

Calinou:
Could we have a separate `cl_maxfps_menu` cvar that defaults to something like 120 (like CS:GO)?

PR #744, opened 2023-02-12
Add cl_maxfps_menu

zigg1zagg1:
Adds the command cl_maxfps_menu, defaults to using the refresh rate
still caps at 30 when minimised

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -1830,7 +1830,13 @@
     },
     "cl_maxfps": {
       "default": "0",
-      "desc": "This variable sets the maximum limit for frames-per-second. Please see vid_vsync, cl_independentphysics, and cl_physfps.",
+      "desc": "This variable sets the maximum limit for frames-per-second while in-game. Please see cl_maxfps_menu, vid_vsync, cl_independentphysics, and cl_physfps.",
+      "group-id": "8",
+      "type": "float"
+    },
+    "cl_maxfps_menu": {
+      "default": "0",
+      "desc": "This variable sets the maximum limit for frames-per-second while disconnected, has a minimum of 30, set lower to use display frequency.",
       "group-id": "8",
       "type": "float"
     },
--- a/src/cl_main.c
+++ b/src/cl_main.c
@@ -143,6 +143,7 @@ static cvar_t cl_sv_packetsync = { "cl_sv_packetsync", "1" };
 cvar_t	cl_sbar		= {"cl_sbar", "0"};
 cvar_t	cl_hudswap	= {"cl_hudswap", "0"};
 cvar_t	cl_maxfps	= {"cl_maxfps", "0"};
+cvar_t	cl_maxfps_menu	= {"cl_maxfps_menu", "0"};
 cvar_t	cl_physfps	= {"cl_physfps", "0"};	//#fps
 cvar_t	cl_physfps_spectator = {"cl_physfps_spectator", "77"};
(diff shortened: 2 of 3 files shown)


ciscon:
can you change the default value to actually be 0 as the help on the cvar specifies?

ciscon:
nevermind it needs a few more fixes as well, without setting anything it just goes into a spin as of now

zigg1zagg1:
Didnt catch that, thanks. I think it makes the most sense to use the refresh rate by default, makes the mouse feel correct without wasting power on super high fps by default.
If its not capping when set to `""` that sounds like its not getting the refresh rate somehow, does vid_displayfrequency show the correct value?

ciscon:
so only the latched value is correct if it's set to 0, i'm doing a quick fix trying to grab the proper value (and if all else fails falling back to 30)

ciscon:
this will fix it all:

```diff --git a/src/cl_main.c b/src/cl_main.c
index fce21bd7..9bdc3842 100644
--- a/src/cl_main.c
+++ b/src/cl_main.c
@@ -143,7 +143,7 @@ static cvar_t cl_sv_packetsync = { "cl_sv_packetsync", "1" };
 cvar_t cl_sbar         = {"cl_sbar", "0"};
 cvar_t cl_hudswap      = {"cl_hudswap", "0"};
 cvar_t cl_maxfps       = {"cl_maxfps", "0"};
-cvar_t cl_maxfps_menu  = {"cl_maxfps_menu", ""};
+cvar_t cl_maxfps_menu  = {"cl_maxfps_menu", "0"};
 cvar_t cl_physfps      = {"cl_physfps", "0"};  //#fps
 cvar_t cl_physfps_spectator = {"cl_physfps_spectator", "77"};
 cvar_t  cl_independentPhysics = {"cl_independentPhysics", "1", 0, Rulesets_OnChange_indphys};
@@ -2115,14 +2115,11 @@ static double CL_MinFrameTime (void)
        if (cls.timedemo || Movie_IsCapturing())
                return 0;
 
-       if (cls.state == ca_disconnected)
-               if (cl_maxfps_menu.value > 0 && cl_maxfps_menu.value >= 30)
+       if ((cls.state == ca_disconnected) || (Minimized && !cls.download))
+               if (cl_maxfps_menu.value >= 30)
                        return 1 / cl_maxfps_menu.value;
                else
-                       return 1 / r_displayRefresh.value;
-
-       if (cls.state == (Minimized && !cls.download))
-               return 1 / 30.0;
+                       return 1 / (r_displayRefresh.autoString ? (uintptr_t)r_displayRefresh.autoString : 30.0);
 
        if (cls.demoplayback)
        {
```

zigg1zagg1:
Thanks, do you think it makes more sense to use 0 as the default? i guess that would be consistent with how it works now

ciscon:
yeah most things that are integers/floats default to 0 which just means auto.  if you still want the minimized/not downloading to default to 30 (though i think refresh/menufps is just fine) you'd want
if (Minimized && !cls.download) return 1/30.0
above this new logic, otherwise it'll never get hit if they are disconnected, though like i said i think this is fine

ciscon:
it also means that if you've minimized it but it's being jammed through a composite layer, you can still be streaming it and it won't look like garbage

zigg1zagg1:
It now defaults to 0, which uses the refresh rate, gets the latched value of displayfrequency and works in windowed mode.

PR #750, opened 2023-03-17
Modernize defaults, and add more settings to menu.

dsvensson:
Adds menu times for video renderer related settings and changes defaults for:

* gl_anisotropy to 16, reduces risk of bad rendering of fence textures at steep angles.
* vid_renderer to 1, the modern renderer, transparent fallback to classic.
* vid_software_palette to 1, a very common recommendation on discord wrt gamma.
* r_dynamic to 2, dynamic lighting on the GPU for modern.

All configuration settings with updated defaults have also gained options menu entries. In addition the basic vid_framebuffer* related settings have gained menu entries.


src/menu_options.c
```diff
@@ -1257,6 +1299,24 @@ setting settsystem_arr[] = {
 	ADDSET_CUSTOM("Fullscreen", FullScreenRead, FullScreenToggle, "Toggle between fullscreen and windowed mode."),
 	ADDSET_ACTION("Apply Changes", VideoApplySettings, "Restarts the renderer and applies the selected resolution."),
 
+#ifdef EZ_MULTIPLE_RENDERERS
+	ADDSET_SEPARATOR("Renderer"),
+	ADDSET_ENUM("Mode", vid_renderer, vid_renderer_enum),
+	ADDSET_ACTION("Apply Changes", RendererRestart, "Restarts the renderer."),
+#endif
```
dsvensson, line 1306: Not really happy with having this in a separate section with apply, but it doesn't really fit into the yes/no in the video settings above and I couldn't think of any better place.

src/r_texture_cvars.c
```diff
@@ -44,7 +44,7 @@ cvar_t gl_picmip = { "gl_picmip", "0", CVAR_RELOAD_GFX };
 cvar_t gl_miptexLevel = { "gl_miptexLevel", "0", CVAR_RELOAD_GFX, OnChange_gl_miptexLevel };
 cvar_t gl_texturemode = { "gl_texturemode", "GL_LINEAR_MIPMAP_LINEAR", 0, OnChange_gl_texturemode };
 cvar_t gl_texturemode_viewmodels = { "gl_texturemode_viewmodels", "GL_LINEAR", 0, OnChange_gl_texturemode };
-cvar_t gl_anisotropy = { "gl_anisotropy","1", 0, OnChange_gl_anisotropy };
+cvar_t gl_anisotropy = { "gl_anisotropy","16", 0, OnChange_gl_anisotropy };
```
dsvensson, line 47: Think this has close to zero performance impact on hardware not from the stone age, and there is a menu item to control it. With low settings both renderers easily filter out for example "metal grate" fence textures at steep rendering angles.

tcsabina:
Nano,
So which setting's default changed? vid_renderer to 1, and gl_anisotropy to 16?

I think we need to update the in-build help for these new settings in the menu, right? There is that .json file (help_something), which is being used if you hit F1 in while you are in the menu. Or am I mistaken?

dsvensson:
Oh right, forgot to update the json file, good catch! Problem is that r_dynamic is either 1 or 2 depending on modern or all build. I'll check if it can be always 2. Think that has the same effect as setting the default to 1 if on classic.

PR #756, opened 2023-03-25
BSP: Fix rounding error in CalcSurfaceExtents.

### LINKED ISSUE #741: BUG: Incorrect lightmap rendering
This bug has been around for quite some time (guess years). Mostly noticed by mappers who has been working around it by for example adjusting texture scale by an invisible amount, or adjusting texture offset. The same maps render correctly in FTE. 

Black bottom part of the texture:
image

Large black parts of the wall:
image

The bottom of the small box has black parts to the left, and light parts to the right:
image

Feels like a precision bug, be it a data type resolution or rounding issue.

According to lurq it might be related to negative texture offsets. Will attach a sample map if I manage to reproduce it with a small map. Have maps with the issues to share without filling up this repo's space.

dsvensson:
The lack of precision frequently caused black artifacts on the map. The workaround used to be adjusting texture scale in the map ever so slightly.

Originally fixed in Quakespasm:

https://github.com/sezero/quakespasm/commit/bfab14040b2f65e02572960f8dc4b1b346f8c151

Fixes #741

### DIFF
--- a/src/r_brushmodel_load.c
+++ b/src/r_brushmodel_load.c
@@ -167,10 +167,24 @@ static void CalcSurfaceExtents(model_t* loadmodel, msurface_t *s) {
 		}
 
 		for (j = 0; j < 2; j++) {
-			val = v->position[0] * tex->vecs[j][0] +
-				v->position[1] * tex->vecs[j][1] +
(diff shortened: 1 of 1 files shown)


tcsabina:
Hi nano,
nice catch! good job!

PR #764, opened 2023-04-16
BUILD: linux - build appimage snapshot

ciscon:


tcsabina:
So @ciscon ,
Should we add this into the 'snapshot' and 'release' build scripts as well? I think that would make sense...

tcsabina (approved):

ciscon:
right, this is why i brought up the deduplication of the build stuff.

Issue #739, opened 2023-01-24
REQ: Add "bindedit" feature


HangTimeQW:
**Is your feature request related to a problem? Please describe.**
Updating binds via console is quite painful as there is no way to edit the existing bind.  This means if you have a bind which is a long string it is time-consuming / error-prone to manually type it in.

**Describe the solution you'd like**
Add a feature equivalent to "aliasedit" that allows in-line editing of existing keybinds.

**Describe alternatives you've considered**
It's possible to work around this by editing configs outside of QW or making all binds call simple aliases, but this is suboptimal.

tcsabina:
@HangTimeQW ,

This feature has been added (thanks to krizej). Could you please test it with a recent snapshot from https://builds.quakeworld.nu/ezquake/snapshots/. Thanks!

ciscon:
fyi, works for me.

tcsabina:
> fyi, works for me.

Alright! I will close this.

PR #768, opened 2023-05-02
add bindedit command

krizej:
#739


krizej:
fd something up i think

PR #770, opened 2023-05-02
Fix player skin texture scaling and filtering

### LINKED ISSUE #687: BUG: Player model is unaffected by gl_texturemode
**ezQuake version:**
3.6-alpha10 

**Describe the bug**
Player models appear to use `gl_linear` texture filtering regardless of what `gl_texturemode` is set to

**To Reproduce**
1. Set `noskins "1"`, `r_enemyskincolor ""` and `r_teamskincolor ""`
2. set `gl_texturemode "gl_nearest"`
3. see players still appearing with smooth texture filtering

krizej:
previously the player model was affected by `gl_texturemode` only if `gl_playermip` was not 0, so if you wanted to use the `GL_NEAREST` filter you had to sacrifice some quality (fix #687)

the issue with scaling is visible below
scaling comparison

### DIFF
--- a/src/skin.c
+++ b/src/skin.c
@@ -759,8 +759,8 @@ void R_TranslatePlayerSkin(int playernum)
 		translate[BOTTOM_RANGE + i] = (bottom < 128) ? (bottom + i) : (bottom + 15 - i);
 	}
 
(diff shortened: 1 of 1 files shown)


tcsabina:
@krizej,

Nice!

Issue #772, opened 2023-05-03
REQ: date/time shown on endgame scoreboard


mushis:
Would be nice to have the date and time shown on the endgame scoreboard.
So when any match is finished, scores are shown as usual, but also the date and time. I would suggest the position to be bottom center, but any is ok.

Nice to have: the date/time format could be customized.
Nice to have: while we're at it, add a hud element for it.

blaps:
Much missed feature. Could also be a setting to print client version after date/time.

Why would you have date in hud? :) time already exists as an element. 
What I really miss tho, is the +showteamscores (what you see when you tab) as a hud element, but that's for a separate ticket.

tcsabina:
Hi @mushis,

This has been added by PR #779. Please get a recent snapshot build from https://builds.quakeworld.nu/ezquake/snapshots/ and test it.

mushis:
checked, working. 
Thanks

PR #773, opened 2023-05-03
date/time hud element for the scoreboard

krizej:
adds #772 

in-game:
- added a new hud element, scoreclock
  - it is only displayed when the scoreboard is visible (either at the end of the game or when using `+showscores`/`+showteamscores`)
  - `hud_scoreclock_format` cvar controls the format of the clock according to [this](https://www.man7.org/linux/man-pages/man3/strftime.3.html)

in-code:
- hud elements are no longer drawn by `HUD_DrawObject` if they have the flag `HUD_NO_DRAW` set (it was not used earlier)

notes:
- I think the `HUD_ON_INTERMISSION` flag does not work

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -7099,7 +7099,7 @@
       "type": "string"
     },
     "hud_scoreclock_format": {
-      "desc": "Controls in what format the scoreclock is displayed.",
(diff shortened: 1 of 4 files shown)


tcsabina:
@krizej,
Nice work!

I will close and right away reopen the PR, so the update pipelines (with the linux appimage build) are triggered as well.

krizej:
i will close this for now and remake it later with some changes

PR #774, opened 2023-05-04
fix z-fighting 

krizej:
[this commit](https://github.com/QW-Group/ezquake-source/commit/155e9d731d419701c1c813b3d5ddef78abedb0ec) broke the z-fighting fix by making the units offset negative to prevent buttons being pushed into walls. On my machine it seems that negative offsets are no different than 0, however a negative factor can offset a brush in the opposing direction. I made the factor customizable (`gl_brush_polygonoffset_factor`, range -1 to 1), so players are free to choose which way the brushes are offset.

tested in the same situation as #747 as well as on a [custom map](https://cdn.discordapp.com/attachments/672796268585680906/1103663973045391360/zfixtest.bsp) with doors like this facing all 4 directions

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -4608,7 +4608,19 @@
       "type": "float",
       "values": [
         {
-          "description": "Value in range 0-25.",
+          "description": "Value in range 0 to 2.",
+          "name": "*"
+        }
+      ]
+    },
+    "gl_brush_polygonoffset_factor": {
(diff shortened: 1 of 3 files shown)


tcsabina:
@krizej, same here: I close and reopen...

tcsabina:
Thanks @krizej for the contribution!

Issue #673, opened 2022-09-01
BUG: gunkick is broken with fast firing weapons


zigg1zagg1:
ezQuake version: 3.6-alpha10

The camera kick when firing weapons breaks when using rapid fire weapons like the sng or lg.
The kick happens normally at first but stops after a few shots, firing the ssg specifically makes the gunkick happen again.

**To Reproduce**
1 set v_gunkick to 1
2 fire sg / rl / gl and notice correct behaviour
3 fire ng / sng / lg and notice the gunkick stops after a while
4 fire ssg to make the gunkick start again

krizej:
seems to be fixed in 3.6.3-dev

zigg1zagg1:
Still happening here, video example:
https://user-images.githubusercontent.com/73136291/236046208-847d473c-b7d6-40a1-ab32-f7e57cba6bba.mp4

notice how when firing the super nailgun the view kick stops after a few shots, and then doesnt work when firing the regular shotgun after, it then fixes when i fire the super shotgun

krizej:
i tested this on linux so this must be either windows specific or caused by some other setting

tcsabina:
@zigg1zagg1,
Which client did you use to record that video?
The bugfix of @krizej is just merged to upstream, so I am wondering if you have used your own build with the content of PR #775?

zigg1zagg1:
It was on the latest ezquake, i ended up sharing my config and it turned out to be a framerate related issue, the bug doesnt happen on 1155 fps but does on lower.

That new fix solved it

tcsabina:
Thanks zigg for the feedback!

PR #775, opened 2023-05-04
fix punchangle staying at ideal_punchangle value

krizej:
#673


tcsabina:
Thanks @krizej for the contribution!

Issue #582, opened 2021-08-22
REQ: Allow for the Quake re-release sound files to pass f_modified checks


ginzberg:
Please ensure that your request is predominantly client-based, and not better suited to being logged with the server or mod aspects of QuakeWorld (e.g. mvdsv or ktx)

**Is your feature request related to a problem? Please describe.**
Through the history of Quake, there have been projects to keep the game modern. One example of this is through higher quality textures via the Quake Revitalization Project. One gap we've always had has been related to sound. It simply is not as viable to recreate the sound files in the same way the community has with the textures. 

With the re-release of Quake, it appears as though we have been provided a bit of a gift. There are 44khz sound files (in the install directory, they are contained within the `$QUAKEDIR/rerelease/id1/pak0.pak`, within the `/sound` directory.) These files can be extracted and repackaged by a user for them to play with ezQuake

**Describe the solution you'd like**
Currently `f_modified` will report the modified sound files. I would like for the criteria that determines whether these sound files have been modified to also allow for the new re-release sounds to pass. This would essentially allow for two valid "originals" as they are not truly custom.

**Describe alternatives you've considered**
Not an alternative solution, but a possible increased scope option would be to add measures for ezQuake to incorporate the re-release pak0.pak file if it is present. This would eliminate the need for users to unpak and move the files locally. It would also limit the risk of the files being redistributed, as some users may not be familiar with unpak and pak tools. 

**Additional context**

**Impact on server or mods**

ginzberg:
![ezquake009](https://user-images.githubusercontent.com/775372/130365260-c7b77755-e869-4745-b72d-59b1f5af40cf.png)

f_modified reports the following modified sound files:
armor1.wav
damage3.wav
health1.wav
inv1.wav
inv2.wav
inv3.wav
itembk2.wav
land.wav
land2.wav
water1.wav
water2.wav
menu1.wav
menu2.wav
menu3.wav
talk.wav
basekey.wav
runeuse.wav

namtsui:
I tested using the quake rerelease on steam on OpenBSD ezquake 3.2.3. f_modified passes. Tests are welcome on the 3.6 alpha version.

1. unpack rerelease/id1/pak0.pak using [qpackman.](https://github.com/bunder/qpakman)
2. unpack original pak0.pak
3. use the sound/ from rerelease instead inside the extracted original
4. pack pak0.pak
5. f_modified should output all models ok

[...]

PR #790, opened 2023-06-29
add airbut1.wav rerelease sha1 to pass f_modified checks on dm2 and e1m2

namtsui:


namtsui:
I missed a rerelease file airbut1.wav. f_modified reports modified: airbut1.wav on dm2 and e1m2 but not dm3. 

I  haven't tested this but I added it in the same manner as the prior changes.

https://github.com/QW-Group/ezquake-source/issues/582

namtsui:
while here also have consistent naming for all airbut1 button sounds.

PR #791, opened 2023-06-30
move the home directory location (linux only)

krizej:
move ~/.ezquake to ~/.local/share/ezquake (or $XDG_DATA_HOME/ezquake if set) to avoid being yet another annoying piece of software that crams itself into your home folder :)

### DIFF
--- a/src/fs.c
+++ b/src/fs.c
@@ -694,7 +694,7 @@ void FS_ShutDown( void ) {
 
 void FS_InitFilesystemEx( qbool guess_cwd ) {
 	int i;
-	char tmp_path[MAX_OSPATH];
+	char tmp_path[MAX_OSPATH], *xdg;
 
 	FS_ShutDown();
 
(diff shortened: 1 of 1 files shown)


tcsabina:
Hi @krizej,

Have you seen #336? That PR also about this topic, and some additional stuff. Could you guys align on this?

src/fs.c
```diff
@@ -761,7 +761,13 @@ void FS_InitFilesystemEx( qbool guess_cwd ) {
 #ifdef _WIN32
 		strlcat(com_homedir, "/ezQuake", sizeof(com_homedir));
 #else
-		strlcat(com_homedir, "/.ezquake", sizeof(com_homedir));
+		xdg = getenv("XDG_DATA_HOME");
+		if(xdg) {
+			strlcpy(com_homedir, xdg, sizeof(com_homedir));
+			strlcat(com_homedir, "/ezquake", sizeof(com_homedir));
+		} else {
+			strlcat(com_homedir, "/.local/share/ezquake", sizeof(com_homedir));
```
tcsabina, line 769: I think many won't like this change. Why the '.local/share' prefix?

tcsabina (changes requested):

src/fs.c
krizej: if XDG_DATA_HOME isn't set, it should default to ~/.local/share

tcsabina: > if XDG_DATA_HOME isn't set, it should default to ~/.local/share

I know ;).
But that is exactly why I commented...

Why do you think we need to move the ezquake home directory from "~/.ezquake" to "~/.local/share/ezquake"?
What makes you think that this is better, compared to the current one?
Is this your personal taste, or some kind of "standard" of some linux distro? If yes, which distro? What about other distro's "standard".

You see my point?


Also, as I have indicated, there is another PR open which is tackling these directories under linux. That PR seems to cover much more on top of the "home directory". So I rather wait on that PR to be finalized, instead of just merging this PR that only touches to home directory of ezquake.

krizej: > Why do you think we need to move the ezquake home directory to '~/.local/share/ezquake'?
> What makes you think that this is better, compared to the current one?

it follows the XDG spec more closely :P

> I rather wait on that PR to be finalized

then have fun waiting, personally i think we will not hear from the author of that pr anytime soon

sorry for the late reply and i guess this pr can be closed then

PR #793, opened 2023-07-12
BSP: Load bspx lighting for maps compiled with -novanilla.

dsvensson:
Saves a couple of MB by removing redundant lightmaps.


inf1niti:
What happens if a player has coloured lighting turned off?  AFAIK a bunch of people play with lits disabled.

dsvensson:
Then they get the experience they deserve, more ugly. If the ruleset disallows fullbright it will be autoenabled.

PR #796, opened 2023-08-02
LIGHTMAPS: Support for DECOUPLEDLM. 

dsvensson:
This BSPX blob allows mappers to decouple lightmaps from textures so that a map can have higher lightmap resolution without scaling textures, something that otherwise risks introducing aliasing artifacts. Another quirk with texture scaling and vanilla lightmaps is that a texture with 0.5 scale next to one with 1.0 scale causes a visual glitch as the lightmap for the two surfaces have differing resolution.

The feature has been supported by ericw light compiler since end of 2022, and is enabled by setting `_world_units_per_luxel` to an integer value. A setting of `16` is close to vanilla resolution, and a lower value increases the lightmap resolution to the cost of increased bsp size. To reduce size the redundant vanilla lightmaps can be omitted with the `-novanilla` flag to the light compiler.

### DIFF
--- a/src/bspfile.h
+++ b/src/bspfile.h
@@ -289,4 +289,11 @@ typedef struct {
 	byte           ambient_level[NUM_AMBIENTS];
 } dleaf_bsp2_t;
 
+typedef struct {
+	unsigned short lmwidth;
+	unsigned short lmheight;
+	int            lightofs;
+	float          vecs[2][4];
+} dlminfo_t;
+
 #endif /* !__BSPFILE_H__ */
--- a/src/gl_model.h
+++ b/src/gl_model.h
@@ -194,6 +194,8 @@ typedef struct msurface_s {
 	short               texturemins[2];
 	short               extents[2];
 
+	short               lmshift;
+
 	// gl lightmap coordinates
 	int                 light_s;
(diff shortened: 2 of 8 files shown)


dsvensson:
Here is a small map split in half where one side is 1.0 scale, and other side is 0.5 scale. Loading in latest release shows the lightmap issue, and loading with this branch shows the result with worldspawn property `_world_units_per_luxel` set to `4`. The new behavior produces a seamless lightmap across the surfaces with different texture scale, and with the very low value of `4` makes the fence texture shadow very clear.

The map contains vanilla lightmaps to demonstrate the old behavior and is at 741kB out of which 391kB is plain texture data, with `-novanilla` to skip old lightmaps it lands at 617kB, going up to a setting of `8` lands at 456kB, and 16 at 413kB.

I haven't observed any performance impact for maps *not* using this feature. The various loops that now use the lmshift field ought to be vastly dominated by the other activity in the loop bodies, and the shader changes should have next to zero impact.

[cube-q1.zip](https://github.com/QW-Group/ezquake-source/files/12248911/cube-q1.zip)

tcsabina (approved):

Issue #792, opened 2023-07-02
BUG: r_fx_geometry incorrect behavior in smaller viewsize's


mushis:
**ezQuake version:**
3.6.2 f7ba65d

**OS/device including version:**
windows 10

**Describe the bug**
`r_fx_geometry` is miscalculated when screen size smaller than 100

**To Reproduce**
```
/r_fx_geometry 1
/viewsize 90
```

**Expected behavior**
nice lines :)

**Screenshots**
![ezquake000](https://github.com/QW-Group/ezquake-source/assets/3011167/603f1960-3679-431e-a144-c3bfcc01fdc4)
This screenshot is with `viewsize 90`

**Additional context**

zigg1zagg1:
Duplicate of https://github.com/QW-Group/ezquake-source/issues/598 no?
It also happens with cl_sbar 1

tcsabina:
Fixed with #800

Issue #711, opened 2022-12-05
REQ: Spec feature - See through walls


blaps:
As a caster/streamer/qtv user, I'd like to get an even better understanding of the action, by seeing players positions as outlines/skeletons through walls, to be able to predict encounters and better follow the flow of the game. I would not want the feature permanently on, but rather have it on a button toggle and only show it in certain situations.

**Describe the solution you'd like**
An on-the-fly toggleable feature (that doesn't require vid_restart). 
It could have options for outline-width, color, gradient, and such. 
For teamplay, it would be helpful if there is a team/enemy color for the outlines.
Could there be a depth/distance setting, so that the feature only shows relevant players (read: being in relative proximity)?

**Describe alternatives you've considered**
Fill color instead of outline, but it might be too visually cluttering.
There should probably be a setting to include player names (scr_autoid) aswell. 

**Additional context**
Deepfake Photoshop Image1 - gradient outlines. 
![skeleton1](https://user-images.githubusercontent.com/8357443/205592613-23ae9fef-2db0-4bba-bb3c-c7ae0b4f14aa.jpg)
Deepfake Photoshop Image2 - single colored outlines.
![skeleton2](https://user-images.githubusercontent.com/8357443/205592616-0a4e9b8a-0f32-4499-9db1-ca9a47d90e9e.jpg)

mushis:
Is there any technical challenge on this one?

I mean, FTE has something similar, not the outline, but the "health bubble"

tcsabina:
Implemented with #800

PR #800, opened 2023-08-05
Outline tweaks and spectator xray

### LINKED ISSUE #598: BUG: r_fx_geometry doesn't take viewsize into account
Reported on discord by fourier
image

krizej:
**changes**:
- `r_fx_geometry` has been merged with `gl_outline`
- added cvars:
  - `gl_outline_color_world` - sets the color of the world outline
  - `gl_outline_color_model` - sets the color of the model outline
  - `gl_outline_color_team` - sets the color of friendly players' outlines
  - `gl_outline_color_enemy` - sets the color of enemy players' outlines
  - `gl_outline_scale_model` - adjusts the scale of the model outline (smackdown/qcon have 0-1 limit, 0-5 on others)
  - `gl_outline_depth_threshold` - adjusts the threshold for drawing world outlines using depth values
  - `gl_outline_use_player_color` - uses the top and bottom colors of players as outline colors
  - `gl_spec_xray` - demo/qtv-only, see players through walls (uses the gl_outline shader and is affected by outline color and scale values) (#711)
  - `gl_spec_xray_distance` - sets the distance from which you can see players through walls
- fixed #598 (and the duplicate, #792)
- tweaked outlining of turb surfaces (shown below)
image


src/glm_aliasmodel.c
```diff
@@ -489,10 +489,19 @@ static void GLM_RenderPreparedEntities(aliasmodel_draw_type_t type)
 	}
 
 	if (type == aliasmodel_draw_std && alias_draw_instructions[aliasmodel_draw_outlines].num_calls) {
+        extern cvar_t gl_outline_color_model, gl_outline_scale_model;
+
```
dsvensson, line 493: Should add .editorconfig or .clang-format to the repo to avoid silly indentation issues like this.

src/glsl/common.glsl
```diff
@@ -68,6 +68,8 @@ layout(std140, binding=EZQ_GL_BINDINGPOINT_FRAMECONSTANTS) uniform GlobalState {
 	float fogMinZ;
 	float fogMaxZ;
 	float padding;
+
+	vec3 camAngles;
```
dsvensson, line 72: Shouldn't padding above be removed as the added vec3 will hit the next even boundary, or it's not shader struct padding, actually something else?

krizej: yes it should i just didnt know std140 rules at the time (lol)

tcsabina:
Hi @krizej,

Could you define 'r_fx_geometry has been rewritten from scratch and merged with gl_outline' a bit more detailed?
Thanks!

krizej:
the entire geometry shader has been rewritten (erased and written in a different, hopefully better way) 
`r_fx_geometry` cvar has been removed and `gl_outline` is now controlling whether the shader is active
`gl_outline 2` is equivalent to `r_fx_geometry 1`
`gl_outline 3` is equivalent to `r_fx_geometry 1; gl_outline 1`
`gl_outline 1` is unchanged (besides the new additions affecting its behaviour)
i have got one more thing left to do, which is to remove the old `gl_outline 2` which is locked behind the default ruleset and `sv_cheats 1`

ciscon:
So a few things, my only tests have been playing demos, during which I see the old world geometry pop up randomly
ezquake022
Also it doesn't appear that gl_outline_color_enemy is doing anything, it's just taking the color from gl_outline_color_models.
One concern I think will probably be raised at some point is that you can set the outline width pretty large making the character a bit more visible, we may be want to drop the max on that a bit
ezquake023

krizej:
will check those things out tommorow, thanks for testing

suddendeathTV:
Hey, few things:
1) Does this eliminate the ability to allow one form and disallow another? E.g. in the past we could allow outlines of world in a ruleset while disallowing outlines of players in that same ruleset. By merging the two, is that no longer possible? This will make this commit more difficult to accept for many.
2) `gl_spec_xray` is amazing but should probably be limited to QTV/MVD rather than spectator. 

Great work though!

krizej:
hi, ruleset rules are applied on individual modes (`gl_outline 1` and `gl_outline 2`), if the player has `gl_outline 3` it acts like 1+2 (if you get what i mean?) - the code doesnt care if its 3, only if it has 1 (or 2) in it 
`gl_outline 1` = `r_fx_geometry 0; gl_outline 1`, 
`gl_outline 2` = `r_fx_geometry 1; gl_outline 0` and 
`gl_outline 3` = `r_fx_geometry 1; gl_outline 1`

the scale was already changed, i limited it to ranges 0-1 in smackdown and qcon (where model outlines are disabled anyway), and other rulesets have the 0-5 limit (altough maybe i should change it to 0-2 or something, except for ruleset default).

as for `gl_spec_xray`, sure, i too thought that should be the case. i'll drop a commit that limits it to qtv soon.

thanks!

suddendeathTV:
Thanks for the clarification!

Great contributions, I reckon it'll get accepted very soon :-)

ciscon:
with the default gl_outline_world_depth_threshold of 0.3 it looks like this:
ezquake030

PR #810, opened 2023-08-21
BUILD: move txt2c x64 debug target to v143 tools like the others

hemostx:
I found the x64 debug target for subproject txt2c was the only target still using VS 2019 Platform tools (v142) instead of v143.
Didn't have 142 anymore and couldn't run that debug build... until now.


dsvensson:
There's also some issue with destination, noticed that at least shaders are dropped in the same directory as sources. Looks nasty in the project view.

Issue #815, opened 2023-08-28
REQ: Port to pcre2


bgermann:
**Is your feature request related to a problem? Please describe.**
ezquake was autoremoved from Debian testing due to its pcre dependency.

**Describe the solution you'd like**
Please port to pcre2.

**Describe alternatives you've considered**
Make pcre optional or port to some other regex library.

**Additional context**
Debian bug report: https://bugs.debian.org/1000035

tcsabina:
Probably closed with #816.
@bgermann , could you check if this would justify ezquake's presence in Debian testing?

bgermann:
Yes, it would. Will there be a release soon?

tcsabina:
@bgermann,
FYI,
https://www.quakeworld.nu/news/1188/ezquake-364-released

PR #816, opened 2023-08-29
PCRE -> PCRE2

fzwoch:
Here is my quick attempt to move to PCRE2.

It did compile on Linux and macOS. Windows changes were made blind, so I guess there some things that I have missed.

Just did quick test whether the sample regex scripting example still works and it did. I'm pretty sure there are some bugs in there. So I guess it needs quite a big of testing.. so step forward..

Windows will need updating of the pre-compiled lib and header files.

### DIFF
--- a/ezQuake.vcxproj
+++ b/ezQuake.vcxproj
@@ -339,7 +339,7 @@
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>Shlwapi.lib;Setupapi.lib;speexdsp.lib;speex.lib;sndfile.lib;freetype.lib;version.lib;imm32.lib;winmm.lib;turbojpeg.lib;SDL2-static.lib;comctl32.lib;wldap32.lib;Winmm.lib;OpenGL32.lib;ws2_32.lib;SDL2-static.lib;jansson.lib;libpng16.lib;zlib.lib;crypt32.lib;libcurl.lib;libexpatMT.lib;pcre2.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>Shlwapi.lib;Setupapi.lib;speexdsp.lib;speex.lib;sndfile.lib;freetype.lib;version.lib;imm32.lib;winmm.lib;turbojpeg.lib;SDL2-static.lib;comctl32.lib;wldap32.lib;Winmm.lib;OpenGL32.lib;ws2_32.lib;SDL2-static.lib;jansson.lib;libpng16.lib;zlib.lib;crypt32.lib;libcurl.lib;libexpatMT.lib;pcre2-8.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <AdditionalOptions>/ignore:4099 /LTCG %(AdditionalOptions)</AdditionalOptions>
       <LinkTimeCodeGeneration>UseFastLinkTimeCodeGeneration</LinkTimeCodeGeneration>
     </Link>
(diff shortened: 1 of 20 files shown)


ciscon:
hopefully a windows person can jam that stuff in there

dsvensson:
Without committing pcre2 binary lib to the repo :)

ciscon:
that'd be optimal, but that's how it currently is

bgermann:
This fixes #815

dsvensson:
Picked over your changes to my vcpkg branch, and needs a definition for PCRE2_CODE_UNIT_WIDTH on Wintendo,

https://github.com/qw-ctf/ezquake-source/actions/runs/6093624078/job/16533588623

dsvensson:
Oh, I see now that you set it to 8 for Makefile builds, I'll add that to vcxproj later.

dsvensson:
For tracking, pcre2_compile error checking should be performed if its return value is NULL, and the error code is then converted to a string message like this. Should probably create some helper function that performs that print to avoid duplicate code.

Two issues atm, integer passed to %s on errors that crash, and secondly the command list is not populated.

https://gist.github.com/dsvensson/8aa1202b8439718a53b04dcd82c24fec

fzwoch:
Also `pcre2_match()` requires `pcre2_match_data`.

src/EX_browser.c
```diff
@@ -3462,7 +3463,7 @@ static info_filter_t* SB_InfoFilter_Parse(int* count)
 
 		if (split != NULL && split[1] && split[1] != '*') {
 			// no regex is fine
-			filter->regex = pcre_compile(split + 1, PCRE_CASELESS, &error, &error_offset, NULL);
+			filter->regex = pcre2_compile((PCRE2_SPTR)(split + 1), PCRE2_ZERO_TERMINATED, PCRE2_CASELESS, &error, &error_offset, NULL);
 			if (filter->regex == NULL) {
 				Con_Printf("Invalid rule definition: %s\n", error);
```
dsvensson, line 3468: You now pass error as an integer Con_Printf's %s when a pointer is expected. This will cause printf to try to access a string at the address of the error code, which is probably an integer at the range of PCRE_COMPILE_BASE+smallnumber, which will cause the application to crash. See gist from yesterday for how to fix the error handling:

https://gist.github.com/dsvensson/8aa1202b8439718a53b04dcd82c24fec

There are a couple of more cases like this that needs fixing.

fzwoch: I tried to just change them from %s to %d for now. If you find more than this one please report them.

src/EX_browser.c
```diff
@@ -3465,7 +3465,7 @@ static info_filter_t* SB_InfoFilter_Parse(int* count)
 			// no regex is fine
 			filter->regex = pcre2_compile((PCRE2_SPTR)(split + 1), PCRE2_ZERO_TERMINATED, PCRE2_CASELESS, &error, &error_offset, NULL);
 			if (filter->regex == NULL) {
-				Con_Printf("Invalid rule definition: %s\n", error);
+				Con_Printf("Invalid rule definition: %d\n", error);
```
dsvensson, line 3468: This is a regression from how pcre1 was integrated. You should instead fetch the error message like in the gist from last night.

https://gist.github.com/dsvensson/8aa1202b8439718a53b04dcd82c24fec

fzwoch: It gives an error code instead of a string. If someone wants to make it prettier, everyone is welcome.

The prints are very inconsistent in this code. Also the whole code in general does some weird if checking all over the place that seems unnecessary.

dsvensson: Always leave code in better shape than you found it.

ezQuake.vcxproj
```diff
@@ -339,7 +339,7 @@
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
-      <AdditionalDependencies>Shlwapi.lib;Setupapi.lib;speexdsp.lib;speex.lib;sndfile.lib;freetype.lib;version.lib;imm32.lib;winmm.lib;turbojpeg.lib;SDL2-static.lib;comctl32.lib;wldap32.lib;Winmm.lib;OpenGL32.lib;ws2_32.lib;SDL2-static.lib;jansson.lib;libpng16.lib;zlib.lib;crypt32.lib;libcurl.lib;libexpatMT.lib;pcre.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>Shlwapi.lib;Setupapi.lib;speexdsp.lib;speex.lib;sndfile.lib;freetype.lib;version.lib;imm32.lib;winmm.lib;turbojpeg.lib;SDL2-static.lib;comctl32.lib;wldap32.lib;Winmm.lib;OpenGL32.lib;ws2_32.lib;SDL2-static.lib;jansson.lib;libpng16.lib;zlib.lib;crypt32.lib;libcurl.lib;libexpatMT.lib;pcre2.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
dsvensson, line 342: Should be `pcre2-8.lib`

tcsabina:
It is building guys! Ready to ship? :)

fzwoch:
> It is building guys! Ready to ship? :)

That's the rule isn't it?

It is fine from my point of view. But I haven't tested it really. So I guess, review it, test it and see what happens :100:

tcsabina:
> > It is building guys! Ready to ship? :)
> 
> That's the rule isn't it?
> 
> It is fine from my point of view. But I haven't tested it really. So I guess, review it, test it and see what happens 💯

Okay, Tuna. let's do it!

Issue #817, opened 2023-08-30
REQ: Remove checked in dependencies in favor of <package manager>


dsvensson:
Today the repo is polluted with:

* src/vstudio-libs
* src/mingw32-libs
* src/minizip
* src/opengl

In an ideal world it would be fine to rewrite history to avoid this mistake, but immutable hashes are :heart: 

Instead of embedding these dependencies ezQuake could instead use modern packaging systems such as Conan[^1] or vcpkg[^2]. The latter was recently more deeply integrated into Visual Studio[^3]. TrenchBroom is one example in the Quake sphere that uses vcpkg, see their [vcpkg.json](https://github.com/TrenchBroom/TrenchBroom/blob/master/vcpkg.json). It deals with static linking[^4], integrates well into CMake[^5] based builds when that time comes, and probably benefits the AppImage build as well.

As it is today the repo will continue to explode in size as dependencies are bumped or introduced. The new Vulkan backend will most likely want newer Vulkan stuff, and perhaps some other libraries, and the current set of libs in the repo are probably all sorts of vulnerable.

All aboard vcpkg? Yay? Nay? Alternatives?

[^1]: https://conan.io
[^2]: https://vcpkg.io
[^3]: https://devblogs.microsoft.com/cppblog/vcpkg-is-now-included-with-visual-studio/
[^4]: https://devblogs.microsoft.com/cppblog/vcpkg-updates-static-linking-is-now-available/
[^5]: https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/cmake-integration

Calinou:
I'd recommend going with vcpkg, though that likely also implies switching to CMake.

PR #819, opened 2023-09-03
BUILD: Vcpkg instead of vendored binary blobs in the repo.

dsvensson:
Initial testing looks fine, CI build time with a warm cache is fine. Will put an end to the repository bloat, and make it easier to update, or introduce new libraries in the future. Will fit even better into a cmake build when that time comes. The latter could also enable reproducible builds on Linux as well if opting to produce a static binary there with known dependency versions. 

In interesting part about the current checked in blobs is that some of them have a 'd' suffix, which may mean they are debug builds, so that would be an improvement as well.

 #817 

Also relevant to #816 PR.


dsvensson:
Ah look at that, the non-Windows build used the repo version of minizip. Could restore those source files, or change those builds to prefer OS version instead. Up for discussion.

ciscon:
that's why i mentioned the makefile/dep/script stuff, that just needs to be changed real quick - though it looks like all the builds fail.

dsvensson:
The windows build failing here is because it lacks my custom github actions manifest, build succeeds in my repo. Just wanted to share the state of things.

dsvensson:
Oh right, there was the GL header stuff on macOS left. Hopefully just include path issue, otherwise it's fairly harmless to ressurect those header files.

dsvensson:
Turns out SDL exposes wrapper headers for OpenGL that works on all platforms, so switched to that.

dsvensson:
Due to the above OpenGL finding I suspect the vcpkg.json should no longer have to require the opengl stuff. Will try building without tonight.

PR #820, opened 2023-09-12
fixing and tweaking bindedit and aliasedit

krizej:
fixed bindedit crashing when used with an unbound key
after using bindedit or aliasedit the cursor now moves to the beginning of the bind/alias value
some additional validity checks in aliasedit


tcsabina:
Thanks szalona!

PR #829, opened 2023-09-23
TEXTURES: Fix competitive issues with fence textures.

### LINKED ISSUE #789: REQ: better support for fence textures in textureless render modes.
# Problem
A number of the textureless modes available in ezquake blur or block the "transparent" pixels in [fence textures](https://quakewiki.org/wiki/Textures#Fence_Textures).  This can be problematic from a competitive gameplay perspective because depending on the client configuration, some players may be at a competitive disadvantage for maps that use fences.  

As a result, QW mappers provide the advice "don't use fence textures".

## Specific Issues (see rendering examples below):  
- high `gl_picmip` values "blur" the fence textures so that the non-transparent pixels "bleed over" the transparent pixels.
- when using `gl_miptexlevel` transparent pixels are rendered as bright-yellow flat/opaque blocks.

### Describe the solution you'd like
The lowest-effort solution would probably be to flag fence textures and ignore them when using texture resolution reducers.  This might look a bit weird or reduce the amount of FPS saved in maps that use fences, but it's probably the easiest solution to implement.

### Describe alternatives you've considered
A more ideal implementation might continue to process the fence textures as normal, but not bleed into areas of the texture that use the transparency signifier/fence-colour.  I'm not fully versed on how this is implemented though, so I'm not sure how possible this would be. 

## Fence Rendering Examples (Per-Mode)
| Mode | Screenshot |
| ------------- | ------------- |
| **Normal**  | ezquake197 |
| $\color{rgb(0,255,0)}{\textsf{r drawflat}}$ | ezquake199 |
| $\color{rgb(0,255,0)}{\textsf{gl textureless}}$ | ezquake198 |
| $\color{red}{\textsf{gl picmip 8}}$ | ezquake194 |
| $\color{red}{\textsf{gl miptexlevel 1}}$ | ezquake196 |
| $\color{red}{\textsf{gl miptexlevel 3}}$ | ezquake200 |

dsvensson:
Aligns the various uglification settings to how gl_textureless deal with fence textures to avoid putting players with gl_miptexlevel > 0, gl_picmip > 0, gl_max_size < 8192 at a competitive disadvantage by keeping fence textures verbatim. Future improvements could be to actually average or pixelate the textures, while keeping the fence alpha intact, but this fixes the immeadiate issue.

Mostly fixes #789, no competitive penalty for the non-normal examples.

### DIFF
--- a/src/r_brushmodel_load.c
+++ b/src/r_brushmodel_load.c
@@ -1665,6 +1665,7 @@ void R_LoadBrushModelTextures(model_t *m)
 		noscale_flag = 0;
 		noscale_flag = (!gl_scaleModelTextures.value && !m->isworldmodel) ? TEX_NOSCALE : noscale_flag;
 		noscale_flag = (!gl_scaleTurbTextures.value  && Mod_IsTurbTextureName(m, tx->name)) ? TEX_NOSCALE : noscale_flag;
+		noscale_flag = (!gl_scaleAlphaTextures.value  && Mod_IsAlphaTextureName(m, tx->name)) ? TEX_NOSCALE : noscale_flag;
 
 		mipTexLevel  = noscale_flag ? 0 : gl_miptexLevel.value;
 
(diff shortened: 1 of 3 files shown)


dsvensson:
The gl_miptexlevel one.. should the first screenshot be pixelized according to current behavior, and then masked to the high resolution form, or is the lower version something people vibe with even if it may result in competitive disadvantage?
image
image

PR #831, opened 2023-09-23
SDL: Default to no keyboard grab on Windows.

### LINKED ISSUE #830: Revisit vid_grab_keyboard defaults/documentation on Windows
`vid_grab_keyboard` used to only affect Linux, or at least not Windows. With the adoption of vcpkg we bumped to a newer version of SDL that added support for `SDL_HINT_GRAB_KEYBOARD` on Windows. This now prohibits Alt+Tab while in windowed mode, and prevents keyboard shortcuts such as Alt+Shift to switch keymap between configured languages in the OS.

Not sure what the right approach is, if the default should change to 0 on Windows to keep the 3.6.3 and older behavior, or if the documentation should clarify what the configuration does in more detail. I lean towards the former, as it was only by luck + ezq source code that I found the [relevant SDL2 issue](https://github.com/libsdl-org/SDL/issues/4013). 

Current doc:

> _If set, grabs keyboard hotkeys.
> May need to be enabled/disabled depending on your window manager."_

dsvensson:
This is the default on Linux, and incidentally used to be the default on Windows as well, until vcpkg got introduced which brought a modern SDL version with support for keyboard grabbing.

Fixes #830

### DIFF
--- a/src/vid_sdl2.c
+++ b/src/vid_sdl2.c
@@ -191,7 +191,7 @@ cvar_t vid_minimize_on_focus_loss = {"vid_minimize_on_focus_loss", CVAR_DEF1, CV
 // TODO: Move the in_* cvars
 cvar_t in_raw                     = {"in_raw",                     "1",       CVAR_ARCHIVE | CVAR_SILENT, in_raw_callback};
 cvar_t in_grab_windowed_mouse     = {"in_grab_windowed_mouse",     "1",       CVAR_ARCHIVE | CVAR_SILENT, in_grab_windowed_mouse_callback};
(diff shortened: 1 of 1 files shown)


ciscon:
This commit is switching it to 1 instead of CVAR_DEF2 which is 0 in linux/bsd and 1 for "other", don't you just want to add windows to that list (or just change the default to 0)?  In which case we can remove CVAR_DEF2 as that's all it's for, turning keyboard grabbing on in windows by default.

dsvensson:
It was a test of reviewers! 😂

dsvensson:
But you are right, not sure what the implications are on macOS. I'll add Windows to the ifdef for now.

PR #832, opened 2023-09-24
Add "opt" to fixbundle locations

fzwoch:
Apparently on Apple Silicon homebrew installs itself to /opt. So we want to copy/fix libraries from there as well.


fzwoch:
The script cannot handle @loader_path etc paths..

PR #842, opened 2023-10-10
INSTALL: Include all of libsndfile's dependencies.

### LINKED ISSUE #836: BUG: macOS bundle creation is broken
Currently macOS bundle creation does not work correctly.

The reason is that sub-dependencies of `libsndfile` are relatively located with `@loader_path` which are not obvious to find.

dsvensson:
Fixes #836

### DIFF
--- a/misc/install/fixbundle.sh
+++ b/misc/install/fixbundle.sh
@@ -54,7 +54,8 @@ if [ ! -d "$FRAMEWORK_DIR" ]; then
 fi
 
 function get_deps {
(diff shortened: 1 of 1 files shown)


dsvensson:
Needs some testing on a macOS toyputer.

dsvensson:
CI brew needs to install one more dep, and to detect future mistakes it should also assert that all bundled artifacts are of the same arch. Other than that, it correctly loads the previously missing libs from within the .app.

PR #851, opened 2023-10-23
BUILD: macos - use matrix for builds, export proper cpu/arch when doi…

ciscon:
…ng builds

fixup

### DIFF
--- a/.github/workflows/scripts/homebrew.sh
+++ b/.github/workflows/scripts/homebrew.sh
@@ -35,6 +35,7 @@ dependencies=(
 
 
 function install_arm64() {
+    export CPU="arm64"
     # Ignore pre-installed Intel packages, always install what we say.
     export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK="true"
     export HOMEBREW_NO_INSTALL_CLEANUP="true"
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/homebrew.sh
```diff
@@ -50,13 +51,18 @@ function install_arm64() {
 }
 
 function install_intel() {
+    export CPU="x86_64"
```
dsvensson, line 54: CPU? Where is that env used?

dsvensson: Aha, in the Makefile

ciscon: righto, everything builds cleaner now, no need for the "just find a binary and name it 'this'" stuff anymore but meh.  ostensibly we could also make this a universal package by updating the run script, which we can think about.

dsvensson: Ah, but then it shouldn't be in the install_* functions, as those only call brew, and brew only install precompiled packages.

.github/workflows/scripts/homebrew.sh
```diff
@@ -66,10 +72,12 @@ function create_bundle() {
 }
 
 function build_intel() {
+    export CPU="x86_64"
     make strip
 }
 
 function build_arm64() {
+    export CPU="arm64"
```
dsvensson, line 80: Shouldn't this be a make argument instead like DARWIN_TARGET?

PR #859, opened 2023-11-02
update server browser sources

namtsui:
run update_sources.bat, move to https and add --no-if-modified-since

server files haven't been updated since they were added, so I ran the update script. --no-if-modified-since helps wget replace the existing files.

@vikpe's https://raw.githubusercontent.com/vikpe/qw-data/main/hubapi/servers_mvdsv_europe.txt and @ciscon's https://nicotinelounge.com/servers.txt could be good sources of servers for future consideration, but for now just use quakeservers.net.


vikpe:
fyi: added https://raw.githubusercontent.com/vikpe/qw-data/main/hubapi/servers_mvdsv_north_america.txt

PR #886, opened 2024-01-13
Allow server browser scrolling when serverinfo open

pkova:
It's annoying that you have to esc out before using the arrow keys / mouse wheel when you browse servers to check who is online. This PR allows you to keep the serverinfo window open when scrolling up or down. Note that the old behavior for scrolling up / down in the individual serverinfo tabs is preserved with ctrl+arrow etc. 

We do lose the old ctrl+arrow behavior where you could go to the end of the server list by doing ctrl+downarrow, but I think it's such a rare use case that it's fine.

### DIFF
--- a/src/EX_browser.c
+++ b/src/EX_browser.c
@@ -2201,46 +2201,6 @@ void Serverinfo_Key(int key)
 		case K_BACKSPACE:
 			Serverinfo_Stop();
 			break;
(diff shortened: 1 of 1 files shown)


tcsabina:
Hi @pkova,

Thanks for the contribution!

PR #890, opened 2024-01-26
Add user defined onchange callbacks for variables.

MeMcCree:
Callbacks are added by using cvar_set_onchangealias command (cvar_set_onchangealias <cvar> <alias>).


tcsabina:
Hi @MeMcCree,

What should be the use-case for such thing?

MeMcCree:
For example mod can set variables on client. For example grenades in tf, on changes in grenade count edit hud elements to match grenades.

PR #891, opened 2024-01-31
FS: Remove last traces of iscompressed in vfs_zip.

### LINKED ISSUE #885: BUG: ezquake hangs on loading gpl maps in gpl maps pk3
Client hangs when loading a gpl map from the pk3, if you explode them everything is fine.  This problem was introduced with the api16 commit:

commit fbcf8a89f57da1902021cfb8a43dd109e9ccfdb7 (HEAD)
Author: @qqshka <ibolsunov@wallarm.com>
Date:   Sun Oct 22 05:21:18 2023 +0400

    SERVER: add support for API16 so we could run latest KTX
    
    Seems like there is some issues, minor testing was done under MacOS,
    had to compile without optimizations since it crashed with optimizations,
    unfortunatelly it doesn't crash without optimizations so was not able to debug it.
    
    Also need to revisit cmodel.c since during "merge" it lost recent sanity checks validations.

dsvensson:
Most of the uses had already been removed, but remained in the case of a seek. As cmodel now uses seeks during checksum, this uncovered the last use case.

As iscompressed wasn't set, the vfsz->defer remained NULL, and thus reading of pk3 content after a seek failed.

The origin of this file is FTE, ideally a newer version is imported but the rest of the FTE VFS guts has evolved since last import.

fixes #885

### DIFF
--- a/src/vfs_zip.c
+++ b/src/vfs_zip.c
@@ -138,7 +138,6 @@ typedef struct {
 
 	//in case we're forced away.
 	zipfile_t *parent;
-	qbool iscompressed;
 	int pos;
 	int length;	//try and optimise some things
 	int index;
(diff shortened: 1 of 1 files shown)


dsvensson:
Does not fix the issue on unfortunately. Not sure what I was thinking. It no longer exhibits the issue with msvc, but broken with mingw and clang on Windows at least.

ciscon:
fixed it all for me at least, no more breaks or warnings.

BLooD-DoG:
same for me, fixed all the warnings on e1m2 map load

PR #900, opened 2024-02-27
CMD: Support exec of http/https urls

osm:
This change extends the `exec` function to support fetching scripts from a remote location via `http` or `https`.

My intention with this addition is to make it easier to share and ensure that everyone on my team is using the same teamplay config.


tcsabina:
Hi!

Could you please provide a description to this PR?

Thanks!

PR #901, opened 2024-03-02
CL_ParseStufftext: Add list of allowed commands to process

osm:
By setting cl_stufftext_allowed_commands, you can now control what commands the server is allowed to execute on the client.


tcsabina:
As discussed heavily in dev-corner, this is not the way to go.
Closing.

PR #905, opened 2024-03-06
Add stufftext whitelist

osm:
By setting `cl_stufftext_allowed_commands`, you can now control what commands the server is allowed to execute on the client.

There is currently one known way to exploit `stuffcmd` to remotely execute code on the client. A POC video is available at: https://www.youtube.com/watch?v=13Oh1960MFs

There is also a convenient function added, called `set_ktx` that sets `cl_stufftext_allowed_commands` to the values required for KTX to work properly. It also adds `tempaliases` to retain almost identical functionality as before.

However, there are two breaking for the user:
1. Demo downloading from the server will not work.
2. Changing the map by typing `dm3` is no longer supported. Instead, users will have to write `votemap dm3`.

In this PR, the default value for `cl_stufftext_allowed_commands` is `""`, effectively turning the whitelist function `off`. We might want to consider setting this to the values defined in `KTX_COMMANDS` so that users are protected from the RCE shown in the video above by default. But at least now, there is a way for users to easily defend themselves by running `set_ktx`.


ciscon:
I was under the impression you were going to go spoike's way and do the whitelisting on actual commands and have that done when they are executed by looking at the cbuf they came from so we don't have a giant list of hardcoded acceptable commands/aliases, I don't see anybody actually using this in its current form which kind of defeats the purpose.

PR #906, opened 2024-03-10
CI: Workaround for build-dep conflicts on macOS.

dsvensson:
Just homebrew noise, nothing related to the actual build.

### DIFF
--- a/.github/workflows/scripts/homebrew.sh
+++ b/.github/workflows/scripts/homebrew.sh
@@ -33,6 +33,15 @@ dependencies=(
     sdl2_mixer
 )
 
+function cleanup_python() {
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/homebrew.sh
```diff
@@ -52,7 +63,11 @@ function install_arm64() {
 
 function install_intel() {
     export CPU="x86_64"
+    export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK="true"
```
dsvensson, line 66: Used in the arm64 case already, avoids wasting time on updating unrelated packages.

tcsabina:
Thanks nano!

PR #907, opened 2024-03-12
BUILD: disable the x11 gamma workaround logic by default in builds as…

ciscon:
… it's no longer needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ else
 
     # Resolve all symbols at link time
     ifeq ($(SYS),Linux)
-    	CFLAGS += -DX11_GAMMA_WORKAROUND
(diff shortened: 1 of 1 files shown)


ciscon:
i'd appreciate it if an nvidia+linux person could check it out real quick just to make sure everything still works fine (my builds have been like this since i had my 780ti so i'm sure it's fine, but just to double check).

BLooD-DoG:
nvidia+linux person here, no issues that I can see. Although my normal config always had vid_gamma_workaround 0 anyway so not sure I'm a relevant example

ciscon:
great, i too have had it disabled forever via cvar and just wanted to make sure we weren't missing something when actually disabling the bits of code via the defines.

PR #921, opened 2024-07-26
BUILD: cast client_time to time_t for time call so we don't break on …

ciscon:
…32 bit architectures


qqshka (approved):

VVD:
Same for mvdsv, plz.

ciscon:
this logic doesn't exist in mvdsv afaik, the time used there is already time_t.

VVD:
time_t can be 32bit and can be 64bit - crypt_rcon require same size for this on both ends.

PR #922, opened 2024-07-30
CL_IDRIVE: rename to CL_SOCD and make default

dusty-qw:
This renames the confusing "cl_iDrive" cvar to "cl_socd" which is more in line with what is used elsewhere. I also made "1" the default value and added a couple more options to match the functionality of wooting and razer keyboards.

cl_socd
0: Legacy behavior (no priority)
1: Last Input Priority
2: Absolute Priority: moveleft
3: Absolute Priority: moveright

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -1714,20 +1714,28 @@
         }
       ]
     },
-    "cl_iDrive": {
-      "default": "0",
-      "desc": "Emulates \"strafe script\". When turned on, you will always change direction instead of stopping when holding opposing movement keys +moveleft and +moveright, or +forward and +back.",
+    "cl_socd": {
+      "default": "1",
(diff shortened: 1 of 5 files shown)


mushis:
currently "idrive" is reported in the f_check as "i"
will it continue to be reported as "i" now that the var is renamed?

dusty-qw:
> currently "idrive" is reported in the f_check as "i" will it continue to be reported as "i" now that the var is renamed?

That has not been changed for now. However, this should probably be revisited soon if the community moves to allow SOCD in tournament rulesets.

marffinn:
like noone was using external apps to have this functionality on, and undetected. But since KB manufacturers implemented this functionality, all of a sudden Ya'll feel a need to reign it down....

PR #923, opened 2024-08-07
Unify all platforms with CMake

dsvensson:
* Same build system as mvdsv and ktx, polished IDE integrations.
* Same build system on all platforms. No more XML diffs for Visual Studio project.
* Generation of Xcode and Visual Studio projects that work with profiler etc.
* Cross-compilation with mingw for i686/x64, including building of all deps.
* Static builds first class on Windows/Linux/macOS, no need to consider what versions are in distros.
* Sandboxed client on macOS, no risk of leaking filesystem content due to exploits.
* A much more capable build system that makes it easier to deal with future improvements.


src/vid.h
```diff
@@ -21,6 +21,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #ifndef EZQUAKE_VID_HEADER
 #define EZQUAKE_VID_HEADER
 
+#undef strlen
```
dsvensson, line 24: wtf... need to revisit this one o_O

src/vid_sdl2.c
```diff
@@ -1381,6 +1381,8 @@ static void VID_SDL_Init(void)
 		return;
 	}
 
+	SDL_SetHint(SDL_HINT_VIDEO_HIGHDPI_DISABLED, "1");
```
dsvensson, line 1392: Improve commit message, needed to avoid highdpi scaling.

dsvensson:
@tcsabina @qqshka @ciscon This is mostly done by now except for updating CI, old CI still uploading stuff. Maybe some readme stuff left, maybe some bikeshedding here and there, probably some stuff I'm blind to. Read it, try it so we can get it merged. msvc LTO is still champ of performance, but at least possible to use mingw again. Could also merge as is and update CI after merge as this PR will not delete any of the old ways of building.

PR #925, opened 2024-08-12
Warnings and fixes

dsvensson:
Drops dead code, fixes memory leaks, buffer overflows etc.


dsvensson:
Various drive-by's from the cmake branch.

PR #928, opened 2024-08-18
STYLE: Fix some c89 violations.

dsvensson:
The wiki specifies that C89 is the target standard. In reality the codebase adheres to GNU89. When actually enforcing this the behavior of MinGW cross-compilation changes, which allows JSON_INTEGER_FORMAT to work as intended.

Later standards automatically enable __USE_MINGW_ANSI_STDIO=1 in MinGW which is the cause of the jansson breakage.

It seems to be fairly common to combine __STDC_FORMAT_MACROS=1 and __USE_MINGW_ANSI_STDIO=0 in projects that disable it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ RMDIR ?= rm -rf
 MKDIR ?= mkdir -p
 JSON2C ?= ./json2c.sh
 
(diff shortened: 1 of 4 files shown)


dsvensson:
In preparation for the return of mingw (cross-)compilation in the CMake branch. If deciding on newer standard at some later point, some care will be needed to get the correct `long long int` formatting for jansson when cross-compiling.

PR #929, opened 2024-08-24
respect $PKG_CONFIG env. variable to make cross-build easier

a-detiste:
https://alioth-lists.debian.net/pipermail/pkg-games-devel/2024-August/066365.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ endif
 
 LSYS := $(shell echo $(SYS) | tr A-Z a-z)
 CC ?= gcc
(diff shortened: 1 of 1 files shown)


dsvensson:
This is likely fixed in https://github.com/QW-Group/ezquake-source/pull/923

tcsabina:
@a-detiste,
Can we check if #923 fixed the reported problem, or we still need this PR?

Thanks!

a-detiste:
Hi, I'll check again after a release. thanks

PR #934, opened 2024-09-07
CVAR: gl_part_bloodtrails

hemostx:
Toggles classic blood trails when QMB trails are used.


hemostx:
With gl_part_trails 1, you used to have a bunch of gl_part_xxx vars to adjust classic/new particles for many effects - but not blood trails on flying gibs. This is that extra option, with a default of 1 (new particles), as they've always been "new" particles since ... fuhquake?

help_variables.json
```diff
@@ -5178,6 +5178,21 @@
       "group-id": "36",
       "type": "boolean"
     },
+    "gl_part_bloodtrails": {
+      "default": "0",
```
dsvensson, line 5182: It's a bit annoying that the cvar declarations aren't generated from the json files. Would have been neat. Update to 1 as you have in the source.

hemostx: Fixed. So much for my attempts at squashing commits...

PR #941, opened 2024-09-15
VX_TRACKER: Honor r_tracker_colorfix for cl_useimagesinfraglog 0

osm:


tcsabina:
Hi @osm,

Can you show us a 'before' and 'after' screenshot of this PR? Thanks!

osm:
Sure thing:
before
after

tcsabina:
Cool, thanks!

This is not toggle, is it? We are not able to switch between the old and new way of coloring the tracker text, right?
Considering our very conservative community ;), am sure someone would prefer to keep to old style.
What do you think?

osm:
Yes, it is possible to toggle. When `r_tracker_colorfix = 0`, it will only color the weapon, not the player names, just like before.

The problem this PR addresses is that `r_tracker_colorfix = 1` is applied when `cl_useimagesinfraglog = 0`, which wasn’t the case before, and I believe this is a bug.

According to the documentation, `r_tracker_colorfix = 1` should behave as follows:
> Both players in fragline will be colored according to good/bad/myfrag/etc values 
https://github.com/QW-Group/ezquake-source/blob/master/help_variables.json#L16742

dusty-qw:
All colorfix does is fix a bug where the first and second player's names were not the same color.

It sounds like your change will make the weapon colored differently regardless of the colorfix value. Is that correct?

tcsabina:
Thanks @osm !

PR #944, opened 2024-10-03
Allow cl_idrive to be set when using ruleset smackdown

osm:


tcsabina:
Hi @osm,

Sorry, but this is a hard 'no'!
There is a reason while the ruleset is there, and there is a reason while the ruleset doesn't allow the change of cl_idrive.

dusty-qw:
I think this PR deserved more discussion before being closed. The issue being discussed over the last couple months is that new keyboards allow idrive to be enabled in the hardware, outside of ezquake. We shouldn't rely on banning idrive in the ruleset, and doing so might tell a story that isn't true.

Per the results of a recent poll, it appears most of the community is in favor of allowing idrive anyway. For tournaments who wish to ban it, we have blaze's detection script which can give you a better idea if a player is using some form of SOCD.

dirtboxqw:
We have active tournaments that rely on smackdown but disallow idrive / socd as it is considered cheating.

Changing this ruleset completely undermines existing ongoing tournaments and should be disallowed completely.

All modern esports are banning socd and quakeworld should absolutely follow this same course of action to uphold the integrity of the game.

suddendeathTV:
> We have active tournaments that rely on smackdown but disallow idrive / socd as it is considered cheating.
> 
> Changing this ruleset completely undermines existing ongoing tournaments and should be disallowed completely.
> 
> All modern esports are banning socd and quakeworld should absolutely follow this same course of action to uphold the integrity of the game.

In fact only Counter-Strike has banned last input priority SOCD from what I can find. The reason for this is counter-strafing which is made much easier with last input priority.

Better examples are games like Slipgate, Quake Champions, etc which all allow it.

That being said, not changing it mid tournament makes sense.

dusty-qw:
> Changing this ruleset completely undermines existing ongoing tournaments and should be disallowed completely.

No it doesn't, because you shouldn't rely on that in the first place. That's what I've been trying to tell you. The ruleset doesn't prevent anyone from using idrive and is misleading in the best case scenario. If you want to prohibit idrive you have to use blaze's detection script.

PR #945, opened 2024-10-12
Add ruleset smackdrive

osm:


tcsabina:
Hi @osm,
I haven't seen the request or need to have a new ruleset? Where did this come from? Based on what discussion?

I am very much about to close this...

osm:
Hi @tcsabina,

This was discussed by fellow community members in the #talk channel on the EQL Discord server the same day this PR was created. Feel free to scroll back and check it out.

The `2024` ruleset was also added to your wiki on August 15th by BPS and has been implemented accordingly. https://github.com/QW-Group/ezquake-source/wiki/Rulesets 

If you'd like, I'd be happy to make a formal request if that is what's needed to make this approved.

suddendeathTV:
> Hi @tcsabina,
> 
> This was discussed by fellow community members in the #talk channel on the EQL Discord server the same day this PR was created. Feel free to scroll back and check it out.
> 
> The `2024` ruleset was also added to your wiki on August 15th by BPS and has been implemented accordingly. https://github.com/QW-Group/ezquake-source/wiki/Rulesets
> 
> If you'd like, I'd be happy to make a formal request if that is what's needed to make this approved.

Hey mate,

I like the addition of new rulesets and it's not a nuisance for Toma to accept. However we do need to establish some rules about naming. I think 2024 is too general and essentially hogs the entire year.

Can you think of a better name? I think what people actually want is smackdown but with `cl_idrive` allowed, so `smacklenient` or `smackdrive` would make more sense to me. What do you think?

suddendeathTV:
> so the only difference between this new "2024" ruleset and "smackdown" is the idrive check. Why do we need a new ruleset for it?
> 
> when making a `/f_ruleset check` it prints that idrive is being used. like this: image
> 
> we should just change how it works, so that the Smackdown ruleset works the same as "default" ruleset. I mean, if the user has ruleset Smackdown, lets allow him to enable idrive, `/f_ruleset check` would print `smackdown +i`. so tournament rules can say that the ruleset must be Smackdown, and Idrive is allowed.
> 
> and it would accomplish the same as this "2024" ruleset. more flexible I would say.

I think `cl_idrive` is write protected in Smackdown. Don't think you can use it unless you're on ruleset `default`.

osm:
Hi!

> Can you think of a better name? I think what people actually want is smackdown but with cl_idrive allowed, so smackdrive or something like that would make more sense to me. What do you think?

I like the name `smackdrive`. If it’s approved by the council, I’ll go ahead and rename it.

> I think `cl_idrive` is write protected in Smackdown. Don't think you can use it unless you're on ruleset `default`.

Yes, it is write protected currently.

I created this PR https://github.com/QW-Group/ezquake-source/pull/944 previously that made it possible to toggle `cl_idrive` when using the `smackdown` ruleset. It worked just as @mushis describes it (showing `+i` when performing a check if enabled), it was however closed with a "hard no" comment :)

mushis:
The problem I have with this approach, is that everytime we want to change anything, we need to create a new ruleset.
This is not great.

Talking about iDrive, I would prefer the approach of the PR you mentioned, which is smarter imo (its my suggestion too).
Maybe it was misunderstood, I don't see absolutely ANY problem about making cl_idrive NOT write protected, i.e. the user can set it, BUT its usage is reported with `+i`.

anyway, it is what it is :)

osm:
Rebased the PR and renamed the ruleset from `2024` to `smackdrive`.

osm:
> The problem I have with this approach, is that everytime we want to change anything, we need to create a new ruleset.
This is not great.

I agree. Ideally, it should be implemented like FTE has done it (or similarly to it), but that would require a bit more effort, and I don't have the time for that right now :/

PR #952, opened 2024-10-20
HUD: add hud_gun[2-8]_frame_hide

osm:
This option allows the gun frame to be toggled based on whether or not you actually have the weapon.
frame-hide-off
frame-hide-on


tcsabina:
Thanks @osm!

PR #959, opened 2024-10-28
RENDERER: Add FXAA option.

dsvensson:
Default disabled, ideally used with vid_framebuffer 2 to avoid adding AA to HUD.

NVIDIA FXAA Shader v3.11, BSD license, imported verbatim.

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -21781,6 +21781,86 @@
       "remarks": "Setting vid_framebuffer_scale overrides custom FBO width/height.",
       "type": "integer"
     },
+    "vid_framebuffer_fxaa": {
+      "default": "0",
+      "desc": "Fast Approximate Anti-Aliasing. Enable vid_framebuffer 2 to avoid filtering HUD. Typically between 1-6.",
+      "group-id": "52",
+      "type": "integer",
+      "values": [
+        {
+          "description": "Disabled.",
+          "name": "0"
+        },
+        {
+          "description": "Medium dither, quality 1 (fastest).",
+          "name": "1"
+        },
+        {
+          "description": "Medium dither, quality 2.",
+          "name": "2"
+        },
+        {
+          "description": "Medium dither, quality 3.",
+          "name": "3"
+        },
+        {
+          "description": "Medium dither, quality 4.",
+          "name": "4"
+        },
+        {
+          "description": "Medium dither, quality 5.",
+          "name": "5"
+        },
+        {
+          "description": "Medium dither, quality 6 (highest quality).",
+          "name": "6"
+        },
+        {
+          "description": "Less dither, quality 1 (fastest).",
+          "name": "7"
+        },
+        {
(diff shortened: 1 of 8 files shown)


dsvensson:
Some samples:

image
image
image
image

help_variables.json
```diff
@@ -21781,6 +21781,86 @@
       "remarks": "Setting vid_framebuffer_scale overrides custom FBO width/height.",
       "type": "integer"
     },
+    "vid_framebuffer_fxaa": {
+      "default": "0",
+      "desc": "Fast Approximate Anti-Aliasing. Enable vid_framebuffer 2 to avoid filtering HUD. Typically between 1-6.",
+      "group-id": "52",
+      "type": "boolean",
```
phylter-qw, line 21788: not an integer?

src/glsl/post_process_screen.fragment.glsl
```diff
@@ -2,6 +2,2079 @@
...
+------------------------------------------------------------------------------
+FXAA3 requires RGBL as input unless the following is set,
+
+  #define FXAA_GREEN_AS_LUMA 1
+
+In which case the engine uses green in place of luma,
+and requires RGB input is in a non-linear colorspace.
+
+RGB should be LDR (low dynamic range).
+Specifically do FXAA after tonemapping.
```
phylter-qw, line 136: He says here it's important to apply FXAA after tone mapping and gamma correction, so the input is in a non-linear color space.

In the shader main() function down below, FXAA is being applied _before_ tone mapping.

I'm not sure where gamma correction is applied. Maybe the color buffer is sRGB so it happens automatically. Does anyone know?

phylter-qw (commented):
Much thanks. Such impress. Couple minor scruples I pointed out.

help_variables.json
dsvensson: thanks 😂

src/glsl/post_process_screen.fragment.glsl
dsvensson: Yep, that's correct. But it's somewhat annoying to do, and I couldn't spot any difference. The gamma can be applied via `vid_software_palette 1` at the end of the shader, but if you're using hw gamma you're not going to be doing it afterwards anyways.

dsvensson: Actually tonemap is only applied to world at least, so Fxaa could move to after that. Could also play with applying software palette before fxaa.

dsvensson: But FxaaPixelShader takes sampler2D arg, so I guess the intended situation is to have a pipeline where the world is rendered to a texture tonemapped, and then a fxaa shader is invoked.

dsvensson: I guess I should add the reasoning here in a comment to be one step ahead of future confusion when the next person read the code.

PR #961, opened 2024-11-09
Support FTE_PEXT_TRANS and FTE_PEXT_COLOURMOD.

dsvensson:
FTE_PEXT_TRANS was broken when implemented long ago. This finishes the implementation, and also adds FTE_PEXT_COLOURMOD which was declared, but only implemented in renderer. The latter is a convenient extension field for mappers to use when dealing with alpha brushes later on. For example using the same glass texture with alpha, but tint it via colormod field.

Broken clients (all ezQ versions) will be demoted to opaque entities based on revision.

The renderer is only wired up to use the colourmod and alpha values, but is otherwise unchanged. Even if it has sorting already to blend transparent faces in a somewhat ok order, there are a number of bugs that may for example skip drawing entities behind a transparent entity. This will be dealt with in a later PR that also adds alpha brush support.

When no entity has modified alpha or colormod, the additional bytes are not written. The PF_EXTRA_PFS will however still write one extra byte due to the moving of PF_ONGROUND and PF_SOLID. This is only a problem for buggy clients that have cl_pext_alpha set to 1, and somehow trick the server side check that would disable it if the client is too old, ie you get what you ask for. 

For not fully connected usecases mvd and outdated qtv clients should remain unaffected by this change as long as no alpha nor colormod is used as it doesn't follow the networked player info packet structure. If for example using translucent spawn markers, these are only shown before demo recording usually begins so demos on maps without alpha entities will be playable in any old client.

### DIFF
--- a/src/cl_ents.c
+++ b/src/cl_ents.c
@@ -558,6 +558,15 @@ void CL_ParseDelta (entity_state_t *from, entity_state_t *to, int bits) {
 	}
 #endif
 
+#ifdef FTE_PEXT_COLOURMOD
+	if ((morebits & U_FTE_COLOURMOD) && (cls.fteprotocolextensions & FTE_PEXT_COLOURMOD))
+	{
+		to->colourmod[0] = MSG_ReadByte();
+		to->colourmod[1] = MSG_ReadByte();
+		to->colourmod[2] = MSG_ReadByte();
+	}
+#endif
+
 #ifdef FTE_PEXT_ENTITYDBL
 	if (morebits & U_FTE_ENTITYDBL) {
 		to->number += 512;
@@ -1048,6 +1057,16 @@ void CL_LinkPacketEntities(void)
 		// set trans, 0 and 255 are both opaque, represented by alpha 0.
 		ent.alpha = state->trans == 255 ? 0.0f : (float)state->trans / 254.0f;
 #endif
+#if defined(FTE_PEXT_COLOURMOD)
+		// colourmod 0 is unset, 32 (1.0) incurs no color change
+		if ((state->colourmod[0] > 0 || state->colourmod[1] > 0 || state->colourmod[2] > 0) &&
+			!(state->colourmod[0] == 32 && state->colourmod[1] == 32 && state->colourmod[2] == 32))
+		{
+			ent.r_modelcolor[0] = ((float) state->colourmod[0] * 8.0f) / 256.0f;
+			ent.r_modelcolor[1] = ((float) state->colourmod[1] * 8.0f) / 256.0f;
+			ent.r_modelcolor[2] = ((float) state->colourmod[2] * 8.0f) / 256.0f;
+			ent.renderfx |= RF_FORCECOLOURMOD;
+		}
 #endif
 
 		if (ent.model->flags & EF_ROTATE)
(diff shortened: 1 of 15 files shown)


src/sv_main.c
```diff
@@ -194,6 +194,10 @@ cvar_t sv_pext_mvdsv_serversideweapon = { "sv_pext_mvdsv_serversideweapon", "1"
 
 cvar_t sv_extlimits = { "sv_extlimits", "2" };
 
+#if defined(FTE_PEXT_TRANS)
+cvar_t sv_pext_ezquake_verfortrans = {"pext_ezquake_verfortrans", "7790", CVAR_NONE};
```
dsvensson, line 198: should be updated to something good

src/sv_ents.c
```diff
@@ -730,6 +762,11 @@ static void SV_WritePlayersToClient (client_t *client, client_frame_t *frame, by
 
 		if (pflags & PF_WEAPONFRAME)
 			MSG_WriteByte (msg, ent->v->weaponframe);
+
+#ifdef FTE_PEXT_TRANS
+		if (pflags & PF_TRANS_Z)
+			MSG_WriteByte (msg, ent->xv.alpha ? bound(1, (byte)(ent->xv.alpha * 254.0f), 254) : 255);
```
dsvensson, line 784: Should actually check >0 && < 1 like the other places, and default to zero I think.

src/sv_ents.c
```diff
@@ -955,6 +992,9 @@ void SV_WriteEntitiesToClient (client_t *client, sizebuf_t *msg, qbool recorder)
 			state->colormap = ent->v->colormap;
 			state->skinnum = ent->v->skin;
 			state->effects = TranslateEffects(ent);
+#ifdef FTE_PEXT_TRANS
+			state->trans = ent->xv.alpha ? bound(1, (byte)(ent->xv.alpha * 254.0f), 254) : 0;
```
dsvensson, line 1020: same as above

src/cl_parse.c
```diff
@@ -1879,6 +1879,17 @@ void CL_ParseStatic (qbool extended)
 	// set trans, 0 and 255 are both opaque, represented by alpha 0.
 	ent->alpha = (es.trans == 0 || es.trans == 255) ? 0.0f : (float)es.trans / 254.0f;
 #endif
+#ifdef FTE_PEXT_COLOURMOD
+	// Only use colourmod if unset, or identity
+	if ((es.colourmod[0] > 0.0f || es.colourmod[1] > 0.0f || es.colourmod[2] > 0.0f) &&
```
dsvensson, line 1884: should not be float comparison

src/sv_ents.c
```diff
@@ -955,6 +1002,12 @@ void SV_WriteEntitiesToClient (client_t *client, sizebuf_t *msg, qbool recorder)
 			state->colormap = ent->v->colormap;
 			state->skinnum = ent->v->skin;
 			state->effects = TranslateEffects(ent);
+#ifdef FTE_PEXT_TRANS
+			if (ent->xv.alpha >= 0.0f && ent->xv.alpha <= 1.0f)
```
dsvensson, line 1030: meh one more > vs >=, should be > and <

src/cl_ents.c
```diff
@@ -1045,8 +1054,19 @@ void CL_LinkPacketEntities(void)
 			}
 		}
 #if defined(FTE_PEXT_TRANS)
-		//set trans
-		ent.alpha = state->trans/255.0;
+		// set trans, 0 and 255 are both opaque, represented by alpha 0.
+		ent.alpha = (state->trans == 0 && state->trans == 255) ? 0.0f : (float)state->trans / 254.0f;
```
dsvensson, line 1058: nice CI comment, it's actually enough with == 255

PR #966, opened 2024-11-18
BUG: if we receive null from glGetString do not attempt to call atoi …

ciscon:
…on it


danielstravito (approved):

dsvensson:
oops.. surprise approve by my work user :see_no_evil:

Issue #963, opened 2024-11-15
BUG: <description>Can't build ezQuake 3.6.5 at Devuan 5


12sunflowers:
Please search for existing issues and check for potential duplicates before filing yours.

**ezQuake version:**
3.6.5

**OS/device including version:**
Devuan 5  Daedalus

**Describe the bug**
```
freeartist-devuan@devuan:~/Downloads/ezquake-source-3.6.5$ ./build-linux.sh
[ERROR] Your dist 'devuan' isn't supported by this script. Exiting.
```
if i change Debian to Devuan at build-linux.sh
```
freeartist-devuan@devuan:~/Downloads/ezquake-source-3.6.5$ 
./build-linux.sh
[ERROR] Your Debian version '5' is too old. Exiting.
```

**Expected behavior**
build ezquake at Devuan

ciscon:
i can add devuan to the list but for now you can just comment out the version if statement and make sure the package names it uses are correct for devuan 5

dsvensson:
This is a N+1 problem, it's better to detect expected binaries and tell the user to install whatever package provides such commands.

dsvensson:
https://distrowatch.com/ :eyes:

12sunflowers:
make changes by myself
```
<------>devuan)
<------><------>if [ $VERSION_ID -gt 0 ] && [ $VERSION_ID -lt 5 ]; then
<------><------><------>error "Your Debian version '$VERSION_ID' is too old. Exiting."
<------><------>fi
<------><------># Includes Debian testing/unstable as they don't provide a VERSION_ID
<------><------>install_check_deb
<------><------>;;
```
but has error

```
freeartist-devuan@devuan:~/Downloads/ezquake-source-3.6.5$ 
./build-linux.sh
[STEP ] Install/check dependecies (packages)...
[INFO ] You might be prompted to input your password as superuser privileges are required.
[INFO ] Updating apt repo list... (running with sudo)
[sudo] password for freeartist-devuan: 
[INFO ] Checking/installing required packages... (running with sudo)
[STEP ] Checking out git submodules...
Reading package lists...
Building dependency tree...
Reading state information...
git is already the newest version (1:2.39.5-0+deb12u1).
git set to manually installed.
ninja-build is already the newest version (1.11.1-2~deb12u1).
ninja-build set to manually installed.
build-essential is already the newest version (12.9).
build-essential set to manually installed.
[...]

PR #967, opened 2024-11-18
CMAKE: Switch OpenGL detection to legacy.

dsvensson:
This causes the executable to linked with libGL etc instead of libOpenGL etc. This only changes the behavior on *nix sytems.


ciscon (approved):

Calinou:
Is this related to https://github.com/QW-Group/ezquake-source/issues/963?

dsvensson:
Yep.

PR #977, opened 2024-12-23
RENDERER: Support vid_framebuffer 2 with classic renderer.

dsvensson:
Using "enable" is only a hint, so will still render black if extension is not available. On macOS only the ARB version is available, enabling both EXT and ARB for good measure.


dsvensson:
Have to check on something not macOS how this behaves.

src/glc_framebuffer.c
```diff
@@ -61,6 +61,11 @@ qbool GLC_CompilePostProcessProgram(void)
 		}
 		if (post_process_flags & POST_PROCESS_3DONLY) {
 			strlcat(included_definitions, "#define EZ_POSTPROCESS_OVERLAY\n", sizeof(included_definitions));
+			if (SDL_GL_ExtensionSupported("GL_ARB_draw_buffers")) {
+				strlcat(included_definitions, "#extension GL_ARB_draw_buffers : enable\n", sizeof(included_definitions));
+			} else if (SDL_GL_ExtensionSupported("GL_EXT_draw_buffers")) {
+				strlcat(included_definitions, "#extension GL_EXT_draw_buffers : enable\n", sizeof(included_definitions));
+			}
```
dsvensson, line 68: This might actually not be needed if OpenGL on macOS is initialized to a higher OpenGL version.

PR #978, opened 2024-12-26
RENDERER: Add support for skywind.

dsvensson:
Feature available in the SP community. Ported mostly verbatim from the IronWail engine. Partial cubemap alpha which gets blended with itself to create the illusion of a moving sky. Quite hypnotizing.

To simplify tuning of the skybox the following commands are included in the port:

* /skywind
* /skywind_save
* /skywind_load
* /skywind_lookdir
* /skywind_rotate

If a config file named $skyboxname_wind.cfg in the same directory as the skybox is found, it's automatically loaded and configures the sky.

Implemented for both modern and classic renderers.

### DIFF
--- a/src/r_brushmodel_sky.c
+++ b/src/r_brushmodel_sky.c
@@ -37,6 +37,20 @@ texture_ref skybox_cubeMap;
 
 qbool r_skyboxloaded;
 
+extern cvar_t r_skywind;
+
+static float skywind_dist;
+static float skywind_yaw;
+static float skywind_pitch;
+static float skywind_period;
+
+static char active_skyname[MAX_QPATH];
+
+#define SKYWIND_CFG			"_wind.cfg"
+
+static void Skywind_Load_f(void);
+static void Skywind_Clear(void);
+
 //A sky texture is 256 * 128, with the right side being a masked overlay
 void R_InitSky (texture_t *mt) {
 	int i, j, p, r, g, b;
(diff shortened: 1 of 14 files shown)


src/r_brushmodel_sky.c
```diff
@@ -258,3 +282,262 @@ void R_ClearSkyTextures(void)
...
+	phase -= floor(phase) + 0.5; // [-0.5, 0.5)
+	wind_dir[0] = dist * cp * sy;
+	wind_dir[1] = dist * sp;
+	wind_dir[2] = -dist * cp * cy;
+	*wind_phase = (float) phase;
+
+	return true;
+}
+
+void R_SkyRegisterCvars(void)
```
dsvensson, line 535: bad name, no cvars in sight

src/r_brushmodel_sky.c
```diff
@@ -208,7 +232,7 @@ static qbool Sky_LoadSkyboxTextures(const char* skyname)
 				fixed_size = size;
 
 				if (i == 0) {
-					skybox_cubeMap = R_CreateCubeMap("***skybox***", size, size, TEX_NOCOMPRESS | TEX_MIPMAP | TEX_NOSCALE);
+					skybox_cubeMap = R_CreateCubeMap("***skybox***", size, size, TEX_NOCOMPRESS | TEX_MIPMAP | TEX_NOSCALE | TEX_ALPHA);
```
dsvensson, line 235: This is mostly the only impact of this feature. Skyboxes will be 4bpp instead of 3bpp going forward. Uses more GPU memory, but people who care about GPU memory likely disable skyboxes, and it's not like it's a huge difference.

PR #980, opened 2025-01-02
MODERN: Unbreak and enable draw call batching.

dsvensson:
The drawcall offset was incorrectly double-offset by both calculating polygonOffset_offset in GLM_DrawWorldModelBatch and then for each iteration further offsetting that base offset by drawcall->polygonOffsetSplit via loop variable.


dsvensson:
@meag nice mystery you left there :)

meag:
I'm sure I left worse than that :)  @dsvensson

PR #1007, opened 2025-01-25
LIGHTMAPS: Support for loading e5bgr9 lightmaps.

dsvensson:
This format was added to ericw-tools spring 2024. Initial implementation does not expose it's full range potential, but it loads, and works, which gives mappers the option to target this format instead of the old rgb8 format.

Ported from QuakeSpasm-Spiked, also implemented in FTE.


dsvensson:
Sample map:
https://a.quake.world/maps/dm2.bsp

PR #1012, opened 2025-01-29
Add more userinfo

osm:
The userinfo string will look something like this:

```
...\*ver\ezQuake v3.6.6-dev 7927~cf8de3304\*rsn\smackdown\...
```


dsvensson:
Updating *client also needs to update these:

If so, also update:

https://github.com/QW-Group/mvdsv/blob/master/src/sv_user.c#L356-L379
https://github.com/fte-team/fteqw/blob/master/engine/server/sv_main.c#L2202-L2237

osm:
I will restore `*client` and add a new `*ver` key instead that includes the version details.

PR #1014, opened 2025-02-01
Allow orange, bright red and black colors

osm:
There was a version of qwcl where you could use black and orange colors, but I haven't been able to figure out which version included them or when they were removed.

The current color selection algorithm works like `x * 16 + 8`, where `x` is a row in the Quake palette (as seen here: https://quakewiki.org/wiki/Quake_palette).

So, `color 0` is the same as `0 * 16 + 8`, which maps perfectly to the white color in the palette. The same goes for the other colors between 0 and 13.

I'm uncertain about the exact shade of orange that was possible to use, but my guess is that it was `14 * 16 + 8 == 232`.

`color 14`:
color14

`color 15`:
color15

Old screenshot that demonstrates the black color:
https://www.quakeworld.nu/w/images/4/4c/RFVSSC1.JPG

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -651,6 +651,18 @@
         {
           "description": "Blue",
           "name": "13"
+        },
+        {
+          "description": "Orange.",
+          "name": "14"
+        },
+        {
+          "description": "Bright red.",
+          "name": "15"
+        },
+        {
+          "description": "Black.",
+          "name": "16"
         }
       ]
     },
@@ -4124,6 +4136,18 @@
         {
           "description": "Blue",
           "name": "13"
+        },
+        {
+          "description": "Orange.",
+          "name": "14"
+        },
+        {
+          "description": "Bright red.",
+          "name": "15"
+        },
+        {
+          "description": "Black.",
+          "name": "16"
         }
       ]
     },
@@ -4236,6 +4260,18 @@
         {
           "description": "Blue",
           "name": "13"
+        },
+        {
+          "description": "Orange.",
+          "name": "14"
+        },
+        {
+          "description": "Bright red.",
+          "name": "15"
+        },
+        {
+          "description": "Black.",
+          "name": "16"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


Calinou:
In the long term, I wonder if arbitrary(-ish) colors could be supported. What's the maximum value `color` can hold? If it can hold at least 250 values or so, then we could probably map values 16 and above to [web safe colors](https://colors.watch/color-tables/216-web-safe-colors) which is a collection of 216 colors (so it can fit within a 8-bit value).

osm:
Quake uses 1 byte to store both the bottom and top colors (4 bits for each), if I read the code correctly. So, supporting more colors would be a breaking change, at least if we want to maintain compatibility with the original Quake engine. However, if we only care about QuakeWorld, we can most likely use all 255 colors available in the palette without breaking anything!

This PR, however, is just an attempt to revive the old colors that were possible to use in that old qwcl version.

osm:
I have finally figured out how more than the regular colors could be used in QuakeWorld back in the day.

The bug exists in the following versions:

- qwcl-1.50
- qwcl-1.55
- qwcl-1.64
- qwcl-2.00
- qwcl-2.10
- glqwcl-1.64
- glqwcl-2.10

The bug was fixed in:

- qwcl-2.20
- glqwcl-2.20

Both the software and GL clients correctly restricted the colors to 0 and 13 when using the `color` command; however, the developers forgot to restrict the `bottomcolor` and `topcolor` cvars.

The behavior in the qwcl and glqwcl clients was, however, a bit different. In the software version, you could only use two more colors, `color 14` (orange) and `color 15` (bright red). In the GL version, you could select from the entire palette.

I have been experimenting with the old clients and captured the following screenshots. The software client displays orange and bright red colors, while the GL client displays black.

The following screenshots displays the orange and bright red colors in the original clients (the entirely blue screenshot is how it is rendered in the fixed version)
qwcl-1 50
qwcl-1 55
qwcl-1 64
qwcl-2 00
qwcl-2 10
qwcl-2 20

And these this is the GL client showcasing the black color, the last screenshot is from the fixed version which just rendered the color as white:
glqwcl-1 64
glqwcl-2 10
glqwcl-2 21

I have updated my pull request to be more aligned with the original bug, so instead of just black and orange, you can now also select the bright red color.

- Orange (color 14, mapped to palette index 232)
- Bright red (color 15, mapped to palette index 248)
- Black (color 16, mapped to palette index 0)

orange
brightred
black

PR #1020, opened 2025-02-13
Allow two exec calls during a game

osm:
After this update, the rulesets that previously restricted exec calls now allows up to two exec calls per game.


tcsabina:
No need. Closomg.

PR #1024, opened 2025-02-19
RENDERER: enable centroid sampling for world geometry

hemostx:
Prevents bright artifacts along BSP splits in dark areas when using MSAA.


hemostx:
Before:
ezquake008
After:
ezquake007

PR #1025, opened 2025-02-20
Restrict play in ruleset smack{down,drive}

osm:
The `play` command was previously restricted only in the `qcon` ruleset. This PR extends the restriction to `smackdown` and `smackdrive` as well.

Additionally, it includes a minor refactor to better align the function with other restriction functions - specifically, ensuring that `play` is only be prevented during an ongoing game and allowed in prewar.

Thanks to gnoffa for pointing out that the `play` command should be restricted!


VVD:
You forgot ruleset thunderdome. :-(

osm:
Fixed it in thunderdome as well now @vvd

VVD:
> Fixed it in thunderdome as well now @VVD

Thanks!

PR #1032, opened 2025-03-15
Add port ping probe

osm:
By setting cl_useportpingprobe to 1, your client will perform cl_portpingprobe_probes number of probes on the target server before establishing the connection.

This means that a connectionless ping will be sent to the target server from different source ports to measure which source port that gives the lowest latency, selecting that port as the source for the connection.

When using a proxy, the proxy server will be probed instead. This feature works for both connect and connectbr.

The new variables are:

- cl_useportpingprobe (default: 0)
- cl_portpingprobe_probes (default: 100)

### DIFF
--- a/src/cl_main.c
+++ b/src/cl_main.c
@@ -808,11 +808,31 @@ void CL_Connect_f (void)
 	qbool proxy;
 	char *connect_addr = NULL;
 	char *server_buf = NULL;
+	portpingprobe_status_t status;
 
 	if (Cmd_Argc() != 2) 
 	{
 		Com_Printf ("Usage: %s <server>\n", Cmd_Argv(0));
 		return;
+	}
+
+	// It's not possible to change the source address when connected to a
+	// server, so we'll inform the user and return early
+	if (UsePortPingProbe() && cls.state != ca_disconnected)
+	{
+		Com_Printf("With cl_useportpingprobe enabled, you must disconnect before switching servers\n");
+		return;
+	}
+
+	// The user has already initiated a port ping probe, so we'll abort that
+	// and wait for the worker thread to stop before we proceed.
+	if (UsePortPingProbe() && (status = NET_GetPortPingProbeStatus()) == PORTPINGPROBE_PROBING)
+	{
+		NET_SetPortPingProbeStatus(PORTPINGPROBE_ABORT);
+		while ((status = NET_GetPortPingProbeStatus()) != PORTPINGPROBE_READY)
+		{
+			Sys_MSleep(10);
+		}
 	}
 
 	// in this part proxy means QWFWD proxy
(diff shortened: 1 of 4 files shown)


src/cl_main.c
```diff
@@ -808,13 +808,33 @@ void CL_Connect_f (void)
...
 	{
 		Com_Printf ("Usage: %s <server>\n", Cmd_Argv(0));
 		return;
 	}
 
+	// It's not possible to change the source port when connected to a
+	// server, so we'll inform the user and return early
+	if (UsePortPingProbe() && cls.state != ca_disconnected)
+	{
+		Com_Printf("With cl_useportpingprobe enabled, you must disconnect before switching servers\n");
```
Calinou, line 823: Could we disconnect the client automatically before connecting to another server when ping probe is enabled?

I suppose this will act different in case the server address we `connect` to can't be found thoughg.

Calinou:
I'm curious what can lead to a lower ping according to the source port used. Has there been some research on this?

src/cl_main.c
osm: Yes, this can be changed. I just thought it would be less intrusive for the user to implement it this way

osm:
I’m not aware of any research on this, unfortunately. This is essentially just a re-implementation of a script that @BLooD-DoG  created, there's also a similair script created by @ciscon that is available [here](https://github.com/ciscon/random/blob/master/quake-scripts/quake-test_client_ports.sh). The topic was recently discussed in #dev-corner in the quakeworld discord server.

As I understand it, some routers along the path might implement caching based on the source port being used. By switching the source port, you might be able to trigger a cache miss and get better routing.

dsvensson:
@Calinou https://grok.com/share/bGVnYWN5_41389c3a-c133-4e2e-8d52-2e689e772c9c is the idea. There's really no guanatee that it will last throughout the match. Would be interesting to hear how people use this. If they repeat the procedure between matches to keep winning the latency game or if it's commonly stable across a game session.

BLooD-DoG:
> @Calinou https://grok.com/share/bGVnYWN5_41389c3a-c133-4e2e-8d52-2e689e772c9c is the idea. There's really no guanatee that it will last throughout the match. Would be interesting to hear how people use this. If they repeat the procedure between matches to keep winning the latency game or if it's commonly stable across a game session.

In my experience, it's stable and lasts through an evening session. There are several accounts on discord, see for example https://discord.com/channels/166866762787192833/179895022366228481/1350498316835623015

There is 1 server where I notice ping might start off at 51, then jump to 66, then 77 all within 20 minutes, but that is a single server and not related to playing with client ports. Besides this one server, cycling source ports to go from 82ms to 72ms real ping (90ms to 77ms in scoreboard) has been very reliable on a nightly basis, and stable within that night. It has happened that my ping goes back up to 90 mid-match, but it is rare and the benefit of the feature is definitely clear :)

src/cl_main.c
osm: I received another request for this change, so I modified the code to automatically disconnect if you are already connected.

PR #1041, opened 2025-04-06
Add ignore_{no_weapon,not_enough_ammo}

osm:
Setting ignore_no_weapon to 1 will suppress the "no weapon." message emitted by KTX. Similarly, ignore_not_enough_ammo will suppress the "not enough ammo." message.

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -13453,6 +13453,38 @@
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


Calinou (approved):
I see this all the time during gameplay with a shotgun, so I approve.

PR #1044, opened 2025-04-09
Initialize results to NULL so it's always safe to free

osm:
This fixes the issue @BLooD-DoG reported earlier today (also previously mentioned by @sopparus, but I couldn't reproduce it at the time), where the client would crash.

The crash could be reproduced by:
1. Enabling port ping probing
2. Connecting to a server
3. Disconnecting from the server
4. Connecting to a server with an invalid hostname
5. Crash

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -1835,7 +1835,7 @@ static int NET_PortPingProbeOrchestrator(void *data)
 {
 	portpingprobe_orchestrator_args_t *args = (portpingprobe_orchestrator_args_t *)data;
 	portpingprobe_status_t status = PORTPINGPROBE_READY;
-	portpingprobe_t *results;
+	portpingprobe_t *results = NULL;
 	SDL_Thread **workers = NULL;
 	struct sockaddr_in addr;
 	netadr_t net_addr;
@@ -1914,11 +1914,7 @@ static int NET_PortPingProbeOrchestrator(void *data)
 	status = PORTPINGPROBE_COMPLETED;
 
 cleanup:
-	if (workers != NULL)
-	{
-		Q_free(workers);
-	}
-
+	Q_free(workers);
 	Q_free(results);
 	Q_free(args->target_addr);
 	Q_free(args->original_addr);


tcsabina:
Hi slime,

Lets clean this up a bit!
1) get rid of goto. That is just a no-go
2) what about moving all this probing code to a new file, instead of just adding it to net.c? Would that make sense, be more clear?

And sorry, I should have reviwed the initial PR where you introduced this...

osm:
Hi,

1. I do not wish to remove the goto code. I think it is a well-established and idiomatic pattern for managing resource cleanup in c. I believe it avoids deeply nested if-statements and duplicated cleanup code, making the code simpler and less error-prone.

2. Regarding moving the probe code to a new file, I can definitely do that, but I think it should be done in a separate PR, as the focus of this relatively small PR is solely to fix the free bug. Also, there are two other pending PRs related to the port ping probe, so it would be nice to have them merged before the code is moved to a separate file.

jite:
> Hi slime,
> 
> Lets clean this up a bit!
> 
> 1. get rid of goto. That is just a no-go
> 2. what about moving all this probing code to a new file, instead of just adding it to net.c? Would that make sense, be more clear?
> 
> And sorry, I should have reviwed the initial PR where you introduced this...

1. What? No, goto is a well established concept of dealing with cleanup. In this case the whole patch could be simplified to initing the results variable to NULL though.

2. I agree with @osm , keep the commits separate.

jite:
> > In this case the whole patch could be simplified to initing the results variable to NULL though.
> 
> Good point - I didn’t think about how the `Q_free` macro actually worked when implementing the fix. That said, I think the more detailed labels in this PR serve a purpose, since the code is now a bit easier to read and follow.
> 
> So if everyone is OK with it, I’d like to keep it as is.

I don’t agree, it’s unnecessary and cluttery for no reason.  free() is guaranteed to work with NULL pointers too so it’s not necessary to check it at all (it’s in the C standard).

PR #1045, opened 2025-04-11
Add cl_window_caption 3

osm:
The new window caption shows:

- When connected: <connected players>/<max players> | <map>
- When disconnected: ezQuake v<x.y.z>

The delimiter between elements can be customized using cl_window_caption_delimiter.

Feature requested by bps

### DIFF
--- a/help_variables.json
+++ b/help_variables.json
@@ -3068,8 +3068,18 @@
         {
           "description": "Fixed: ezQuake",
           "name": "2"
+        },
+        {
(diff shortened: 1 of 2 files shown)


hemostx:
Conversely, the menu option for this is already insufficient (only accounts for 0/1, which is *not* "on/off" to begin with) and would need adjusting.

image

(I'm sure Carapace and BD will be fighting over which one of them provides the patch)

Issue #1048, opened 2025-05-15
BUG: vid_framebuffer_scale and vid_framebuffer_multisample break gl_outline functionality


nicocinqw:
**ezQuake version:**
ezQuake 3.6.7-dev 7982~743E84E52
(Specify the Git commit hash if using a non-stable build - this is available through the `/version` command)

**OS/device including version:**
Debian 12/RX 6700XT/ vid_renderer 1

**Describe the bug**
when using vid_framebuffer_scale values below 1 and/or vid_framebuffer_multisample values above 0, outlines shrink in size before disappearing entirely.

**To Reproduce**
1. Load any map
2. Enable gl_outline, any value will work but 2 or 3 are easier to test with as they don't require other players
3. execute vid_framebuffer_multisample 2 then vid_restart
4. Outlines disappear
5. Reset the vid_framebuffer_multisample to 0 then vid_restart again, outlines come back
6. execute vid_framebuffer_scale 0.5
7. Outlines become very very thin

**Expected behavior**
The expected behavior is for the outlines to stay despite the multisample or scale settings

**Screenshots**

gl_outline 3, vid_framebuffer_scale 1, vid_framebuffer_multisample 0
![Image](https://github.com/user-attachments/assets/48d5a9bc-2fff-496f-b5ed-92ec38418742)

gl_outline 3, vid_framebuffer_scale 0.5, vid_framebuffer_multisample 0
![Image](https://github.com/user-attachments/assets/87086206-7b0d-4829-a611-560fb347d6ae)

gl_outline 3, vid_framebuffer_scale 1, vid_framebuffer_multisample 8
![Image](https://github.com/user-attachments/assets/65e1d579-32f7-4ff1-97b5-0691fede66df)

config
https://pastebin.com/UD5qhHv7

**Additional context**
When describing the issue in the NA QuakeWorld discord, Ciscon mentioned that this behavior is similar to why the default sshot mode was changed. Hopefully that is not leading into the wrong direction!

Calinou:
I believe the physical outline size depends on viewport resolution, as it's set in pixels. Changing this would likely break existing configurations, but I agree we should make them resolution-independent in the long run.

It's strange that MSAA makes them smaller though (unless it's 16× MSAA, as graphics drivers actually implement it as 4× SSAA + 4× MSAA). The SSAA part effectively increases the framebuffer resolution, which leads to smaller resolution-dependent elements.

PR #1055, opened 2025-06-12
OUTLINE: Maintain world outline thickness when scaling framebuffer

hemostx:
Actually put a dead parameter in the shader to work.
Scaling factor is computed proportionally to the framebuffer and passed to the shader
Outline conserves overall thickness when supersampling, doesn't get extra fat in low res.

Adresses the "outlines get thinner with vid_framebufferscale 0.x" part of https://github.com/QW-Group/ezquake-source/issues/1048


hemostx:
Oh, wth. The PR wants to pass my previous two already merged commits along with this.

hemostx:
There we go. Cleaned up history.

PR #1072, opened 2025-10-06
iDrive 2

b1az3:
iDrive 2 is designed for players who previously used the Last Input Priority SOCD cleaning behavior. Unlike that mode, iDrive 2 inserts a neutral frame between opposing directional inputs. This results in smoother player movement and slightly reduces the speed gain compared to iDrive 1. While iDrive 2 still provides a competitive advantage, the effect is less noticeable.


b1az3:
removing wrong PR

Calinou:
- Superseded by https://github.com/QW-Group/ezquake-source/pull/1073.

Issue #1078, opened 2025-11-05
Spec xray may disappear ents in actual gameplay


dsvensson:
Video [here](https://discord.com/channels/166866762787192833/179895022366228481/1435545381806280724), shows how player disappears based on `gl_spec_xray_distance` changes. Demo can be downloaded [here](https://discord.com/channels/166866762787192833/179895022366228481/1435393905897115689).

A plausible explanation for this is that the `continue` happens here which would avoid `CL_AddEntity` for that vis frame further down that function.
https://github.com/QW-Group/ezquake-source/blob/0c091de44e38b79fa12d9c406fdc0a7ba777f9ff/src/cl_ents.c#L2067-L2078

`if (cl.spectator || cl.demoplayback || cl.mvdplayback) { ... }` should probably surround that whole code. No need to do that `PM_TraceLine` there when spawned in-game - unless I guess you want adjustable xray in-game, maybe written like that to keep the airmain-diff small, perhaps @redzimski has some fun anecdotes here.

PR #1079, opened 2025-11-05
Fix hide in plain sight bug

osm:
Previously, the xray visibility check was executed for all players. Since the code compares each players distance to gl_spec_xray_distance and skips rendering when distance > gl_spec_xray_distance, players beyond that range were completely omitted from rendering.

As a result, players located farther than the xray distance limit could disappear entirely, effectively allowing them to hide in plain sight.

The check now only runs when gl_spec_xray is enabled and the client is playing a demo.

This should fix: https://github.com/QW-Group/ezquake-source/issues/1078

### DIFF
--- a/src/cl_ents.c
+++ b/src/cl_ents.c
@@ -1800,7 +1800,8 @@ static void CL_LinkPlayers(void)
 	frame_t *frame;
 	trace_t trace;
 	customlight_t cst_lt = {0};
-	extern cvar_t cl_debug_antilag_ghost, cl_debug_antilag_view, gl_spec_xray_distance;
+	extern cvar_t cl_debug_antilag_ghost, cl_debug_antilag_view;
+	extern cvar_t gl_spec_xray, gl_spec_xray_distance;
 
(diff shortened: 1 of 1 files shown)


src/cl_ents.c
```diff
@@ -2062,20 +2063,23 @@ static void CL_LinkPlayers(void)
 			}
 		}
 
-		VectorCopy(cent->lerp_origin, end);
-		end[2] += 12;
-		trace = PM_TraceLine(r_refdef.vieworg, end);
+		if (gl_spec_xray.value && (cls.demoplayback || cls.mvdplayback))
```
dsvensson: This only fixes the issue with gameplay. While it's the most acute, doing something like what `PF_tracebox` does would probably be the real fix, unless I'm misinterpreting what it does.. what I mean is that instead tracing the 8 corners of the entity, and as it's only used outside of actual gameplay, the added cost is irrelevant, and it's probably very cheap. Doing so also makes the `end[2]` tweak redundant.

PR #1090, opened 2026-01-12
TEAMSAYS: Add FPD_ENABLE_PLAYER_COUNT

osm:
The `%e`, `%E`, `%o`, `%O`, and `%s` macros can be abused to count team mates and enemies that are not in sight, effectively allowing players to determine whether enemies are nearby even if they can't see them.

### DIFF
--- a/src/teamplay.c
+++ b/src/teamplay.c
@@ -33,6 +33,15 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #include "qsound.h"
 #include "tp_msgs.h"
 
+#define CASE_PLAYER_COUNT_ALLOWED(c, fn) \
+	case c: \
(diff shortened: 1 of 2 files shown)


HangTimeQW:
> Why not fix the player count macros to only show in-view players at the time of reporting?

I'm unsure how easy it is to count in-view players, e.g. does it matter what fov a player has (perhaps this has already been solved for some other use case, idk).  Also I think it could lead to confusion whereby a player sees an enem0y, but by the time they press the button to call the teammsg, the enemy has momentarily disappeared from view so the bind would report 0 enemies, even if the player intended to report 1.

dsvensson:
Would have to frustum cull and then test four points per player for occlusion, and I guess disable fov changes after death lol.

But something else perhaps more interesting is controlling this with FPD flags just like the rest of the proxy features. If its a feature to enable such reporting, and no servers set that, its compatible with rollout.

dsvensson (approved):

PR #1108, opened 2026-03-01
VX_TRACKER: Add r_tracker_reverse cvar

Xerialen:
## Summary
- Adds `r_tracker_reverse` cvar (default `0`) that reverses the display order of tracker messages so the most recent frag appears at the top instead of the bottom
- Changes `break` to `continue` for expired/empty tracker entries when reversed, so gaps don't cut off remaining messages
- Includes documentation in `help_variables.json`

## Test plan
- Set `r_tracker_reverse 0` (default) — tracker messages appear bottom-to-top as before
- Set `r_tracker_reverse 1` — most recent frag appears at the top of the tracker list

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/hud-modernization-plan.md
+++ b/docs/hud-modernization-plan.md
@@ -1,0 +1,172 @@
+# ezQuake HUD Modernization Plan
+
+## Overview
+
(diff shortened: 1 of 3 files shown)


tcsabina:
@Xerialen,
thanks for the contribution.
And sorry to let this PR idling...

Cheers!

PR #1110, opened 2026-03-23
CVAR: cl_fakeshaft - default to 1

ciscon:


vikpe:
any particular reason why?

ciscon:
i'm tired of telling people they have to change it when they don't use nquake to install it.  everything has hitscan antilag so both 0 and 2 are entirely wrong.

vikpe (approved):

PR #1113, opened 2026-04-09
DEMO: Add spawn crossing warning for demo playback

osm:
Example:
demo_spawnwarn

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -619,6 +619,7 @@ set(client_headers
         ${SOURCE_DIR}/screen.h
         ${SOURCE_DIR}/settings.h
         ${SOURCE_DIR}/settings_page.h
+        ${SOURCE_DIR}/demo_spawnwarn.h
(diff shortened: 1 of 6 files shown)


Calinou:
Looks good, but could this also be allowed while in warmup when connected to a server? (That said, maybe it's KTX's responsibility to implement this.)

This evening, I was checking where it's safe to stand on maps like `steam` to camp a teleporter exit. I needed to have someone take the teleporter to check whether it telefragged me.

osm:
> Looks good, but could this also be allowed while in warmup when connected to a server? (That said, maybe it's KTX's responsibility to implement this.)
> 
> This evening, I was checking where it's safe to stand on maps like `steam` to camp a teleporter exit. I needed to have someone take the teleporter to check whether it telefragged me.

There is a KTX feature that almost does this, called `spawnicide`. But instead of printing that you are crossing a spawn point, it instantly kills you. The feature can be toggled to work only in prewar, but also during games. It is kind of fun for 2on2 on dm2.

PR #1119, opened 2026-05-01
DEMO: Add zip appendix support

osm:
Add demo voice playback using tracks loaded from an appended ZIP. The zip may contain an autoexec.cfg, executed through a restricted command buffer. Currently only demo_voice_load is allowed there.

Works for demo formats that can carry the appended zip data, including mvd and qwd.

New commands and cvars:
- demo_voice_info
- demo_voice_load
- demo_voice_unload
- demo_voice_volume
- demo_voice_enable
- demo_voice_team_filter

Example:
```
$ cat autoexec.cfg
demo_voice_load lethalwiz.ogg offset 0 name ToT_lethalwiz team tot
demo_voice_load oddjob.ogg offset 0 name ToT_Oddjob team tot
demo_voice_load sassa.ogg offset 0 name ToT_sassa team tot
demo_voice_load slime.ogg offset 0 name ToT_slime team tot
$ zip mm3.zip autoexec.cfg lethalwiz.ogg oddjob.ogg sassa.ogg slime.ogg
$ cat mm3.zip >>demo.mvd
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -592,6 +592,7 @@ set(client_headers
         ${SOURCE_DIR}/config_manager.h
         ${SOURCE_DIR}/console.h
         ${SOURCE_DIR}/demo_controls.h
+        ${SOURCE_DIR}/demo_extension_voice.h
         ${SOURCE_DIR}/document_rendering.h
         ${SOURCE_DIR}/ez_button.h
         ${SOURCE_DIR}/ez_controls.h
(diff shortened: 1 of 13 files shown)


dsvensson:
This sounds like a demo-limited version of FTE's _"zip file system at end of file"_ with a couple of audio-playback related functions. Would be a much better fit to make it generic to allow mappers to embed a default .loc, .bot, skybox or whatever as an embedded filesystem at end of file.

osm:
Good point, I was not aware of FTE’s zip fs. Given that, I agree that a generic appended zip fs sounds like a good long term direction.

However, I do not plan to port that as part of this change, but the current implementation should be compatible with that future path: it appends a normal zip archive directly after the demo data, without any custom header or container format. So if/when someone decides to port FTE’s zip fs, the demo-specific zip reader can be removed and the appended archive can instead be mounted through the normal vfs layer.

PR #1140, opened 2026-06-11
Fix r_drawflat destroying alpha cutout on brush-model fence surfaces

dsvensson:
With `r_drawflat 1`, fence/`{`-textured surfaces on inline brush-model entities (e.g. `func_illusionary` palm-tree leaves) rendered as solid flat-coloured rectangles instead of keeping their alpha-cutout silhouette.

Inline submodels inherit `isworldmodel` from the world model, so the renderers enabled the drawflat chains for them and routed their alpha-tested surfaces down paths that drop the alpha test:

- **GLM**: the surfaces went into `drawflat_chain`, drawn with the non-alpha-tested program. Fix: exclude `SURF_DRAWALPHA` surfaces from the drawflat chain (mirroring the world path); they fall through to the texture chain, which still applies the drawflat tint but keeps the silhouette.
- **GLC**: the surfaces were chained onto the global `alphachain`, which `GLC_DrawBrushModel` never draws, so they were missing entirely regardless of `r_drawflat`. Fix: chain them onto the per-texture `texturechain` and flag `alphapass_todo` so the existing alpha pass renders them.

`isworldmodel` inheritance is intentionally untouched — it is what makes the flat tint apply to brush entities at all.

Tested in both renderers: `r_drawflat 0` unchanged; `r_drawflat 1` gives single colour + lightmap with intact leaf silhouettes; GLC now draws the previously missing surfaces.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/glc_brushmodel.c
+++ b/src/glc_brushmodel.c
@@ -1514,7 +1514,6 @@ void GLC_ChainBrushModelSurfaces(model_t* clmodel, entity_t* ent)
 	qbool drawFlatFloors = clmodel->isworldmodel && (r_drawflat.integer == 2 || r_drawflat.integer == 1) && r_drawflat_mode.integer == 0;
 	qbool drawFlatWalls = clmodel->isworldmodel && (r_drawflat.integer == 3 || r_drawflat.integer == 1) && r_drawflat_mode.integer == 0;
 	extern msurface_t* skychain;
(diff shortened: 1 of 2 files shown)


dsvensson:
palm tree with some of the leafs as func_illusionary with r_drawflat 1 before fix:
image

